Issue #83, opened 2021-07-12
Add a validation error when user uses an ExistingDimension without providing a column.output_uri_template.


canwaf:
We can't currently cope with this (1). Make sure to add a validation test for this.

robons:
Found the stuff I was using to explain this the other day!

```python
import pandas as pd

data = pd.DataFrame({
    "Some Dimension": ["A", "B", "C"],
    "ObsVal": [1,2,3]
})


"""
When defining the column the user must define the `output_uri_template` otherwise we have no idea how to map each cell value to the appropriate code-list concept. 

(This is true when the user is offline/air-gapped or when the dimension/codelist are defined somewhere other than on PMD). 
"""

QbColumn("Some Dimension", ExistingQbDimension("http://some/dimension/uri"),
            output_uri_template="http://some/code/list/{+some_dimension}")

"""
  We need to know what the output_uri_template is so we know how to map the CSV-W to RDF.

    {
        "titles": "Some Dimension",
        "name": "some_dimension",
        "propertyUrl": "http://some/dimension/uri",
        "valueUrl": "<output_uri_template>"
    }

                                          
    :observation/1 <http://some/dimension/uri> <http://some-concept-scheme/concept/some-value>.

   How we can lookup the codelist's URI once we know where the dimension is defined:

    :dataSet a qb:DataSet;
             qb:structure :dsd.
             
    :dsd a qb:DataStructureDefinition;
         qb:component :dimensionComponent. 
    
    :dimensionComponent a qb:ComponentSpecification;
                        qb:dimension ?dimension.

    ?dimension qb:codeList ?codeListUri.
"""
```

PR #92, opened 2021-07-15
Issue#83 - Add a validation error when user uses an ExistingDimension without providing a column.output_uri_template.

santhosh-thangavel:
1.QB with an existing dimension and without an output_uri_template is defined.
2.A Validation error is generated when user provides an existing dimension but not a column.output_uri_template.
3.Test case written to capture the number of errors.

#83

### DIFF
--- a/csvqb/csvqb/tests/unit/cube/qb/errorvalidation.py
+++ b/csvqb/csvqb/tests/unit/cube/qb/errorvalidation.py
@@ -1,7 +1,6 @@
 import unittest
 import pandas as pd
 
-
 from csvqb.models.cube import *
 from csvqb.models.rdf import URI
 from csvqb.utils.qb.cube import validate_qb_component_constraints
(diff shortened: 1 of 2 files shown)


santhosh-thangavel:
@canwaf Please review my pull request. Thank you

robons (approved):

csvqb/csvqb/tests/unit/cube/qb/errorvalidation.py
```diff
@@ -61,6 +62,28 @@ def test_multi_measure_qb_definition(self):
...
+            QbColumn("Existing Dimension", ExistingQbDimension("http://example.org/dimensions/location")),
+            QbColumn("Value",
+                     QbSingleMeasureObservationValue(ExistingQbUnit("http://some/unit"),
+                                                     ExistingQbMeasure("http://some/measure")))
+        ])
+
+        errors = cube.validate()
+        errors += validate_qb_component_constraints(cube)
+
+        self.assertEqual(1, len(errors))
```
robons, line 85: I think we should be more specific about the error we expect - i.e. check that the text matches to some degree. Otherwise it could be another error we're triggering and we'd never know.

santhosh-thangavel: Expected error message is asserted and the error message matches to what we are trying to test.

canwaf (approved):
Looks logical to me.

csvqb/csvqb/tests/unit/cube/qb/errorvalidation.py
robons: Looks good to me. Thanks!

Issue #62, opened 2021-07-06
Test cube validation


robons:
* implement validation method in the `CubeMetadata` model.
  * validation tests (do **not** use the info.json configuration to set these tests up, use the internal common model). Use the `cube.validate()` method for the following tests.
    * Define a cube with two columns with the same title. Test that a suitable error message is generated on validation.
    * Test that defining a column with a title that does not exist in the dataframe provided generates a suitable error message on validation.
    * Test that neglecting to define a column mapping for a column in your input dataframe causes a suitable error to be raised.

PR #111, opened 2021-07-29
Issue#62 Test cube validation 

santhosh-thangavel:
Suitable error message is generated when
1.  A cube with two columns with the same title is defined.
2. Defining a column with a title that doesn't exist in the DataFrame.
3.  A column in input DataFrame doesn't have a column mapping.

#62


santhosh-thangavel:
@canwaf please, review my PR.

canwaf (approved):
I'm approve these changes; however there is potential for a bug not in your implementation but in the surrounding code. I'll raise an issue separately.

Issue #86, opened 2021-07-13
Test `NewQbDimension.from_data`


robons:
Test `NewQbDimension.from_data` correctly extracts and creates a valid `NewQbCodeList` and valid `NewQbConcept`s for each of the distinct values in the data provided.

It's probably best to create a new class for testing DSD components in the `tests.unit.cube.qb` package.

PR #116, opened 2021-08-03
Issue#86 - Test `NewQbDimension.from_data

santhosh-thangavel:
1. Test case is added to check if NewQbDimension correctly extracts and creates a Valid NewQbCodeList & NewQbConcept
2.Unique concept labels are asserted
3.Asserted NewQbDimension concepts does not have a parent
4. Asserted to see if concept labels are converted into something that can be used in a URI path segment (pathify).

#86

### DIFF
--- a/csvqb/csvqb/tests/unit/cube/qb/test_qb_dimension.py
+++ b/csvqb/csvqb/tests/unit/cube/qb/test_qb_dimension.py
@@ -38,6 +38,6 @@ def test_newqbdimension_extracts_newqbcodelist_newqbconcept():
     code = [c.code for c in new_dimension.code_list.concepts]
     assert "a" in code
     assert "c" in code
-    assert "c" in code
+    assert "d" in code
     assert "e" in code
     assert "g" in code
--- a/csvqb/csvqb/tests/unit/cube/qb/test_qb_dimension.py
+++ b/csvqb/csvqb/tests/unit/cube/qb/test_qb_dimension.py
@@ -30,8 +30,13 @@ def test_newqbdimension_extracts_newqbcodelist_newqbconcept():
     assert "E" in concept_labels
     assert "G" in concept_labels
 
-    # assert parent_code is None
-    assert new_dimension.code_list.concepts[0].parent_code is None
+    # assert parent is None for distinct list of concepts
+    parent_code_var = [c.parent_code for c in new_dimension.code_list.concepts]
+    assert parent_code_var[0] is None
+    assert parent_code_var[1] is None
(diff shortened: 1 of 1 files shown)


santhosh-thangavel:
@canwaf Please, review my pull request.

csvqb/csvqb/tests/unit/cube/qb/test_qb_dimension.py
```diff
@@ -0,0 +1,42 @@
...
+    # List of concept labels are asserted Individually
+    concept_labels = [c.label for c in new_dimension.code_list.concepts]
+    assert "A" in concept_labels
+    assert "C" in concept_labels
+    assert "D" in concept_labels
+    assert "E" in concept_labels
+    assert "G" in concept_labels
+
+    # assert parent_code is None
+    assert new_dimension.code_list.concepts[0].parent_code is None
```
robons, line 34: It would be a good idea if we could assert this for *all* concepts in the codelist.

santhosh-thangavel: asserted parent is None for list of distinct concepts.

csvqb/csvqb/tests/unit/cube/qb/test_qb_dimension.py
```diff
@@ -0,0 +1,47 @@
...
+    concept_labels = [c.label for c in new_dimension.code_list.concepts]
+    assert "A" in concept_labels
+    assert "C" in concept_labels
+    assert "D" in concept_labels
+    assert "E" in concept_labels
+    assert "G" in concept_labels
+
+    # assert parent is None for distinct list of concepts
+    parent_code_var = [c.parent_code for c in new_dimension.code_list.concepts]
+    assert parent_code_var[0] is None
```
robons, line 35: There's a nicer way to do this using the [all](https://docs.python.org/3/library/functions.html#all) function with a list comprehension.

Want to give that a go?

santhosh-thangavel: Sure

santhosh-thangavel:
@canwaf Please, review my pull request.

csvqb/csvqb/tests/unit/cube/qb/test_qb_dimension.py
santhosh-thangavel: Used [any](https://www.w3.org/2009/08/skos-reference/skos.html#hasTopConcept) function with a list comprehension

canwaf (approved):

PR #121, opened 2021-08-05
Charles Rendle/issue87

CharlesRendle:
Adds validation error to `ExistingQbAttribute` when a `output_uri_template` is not defined and a related unit test.

Removes references to `qb:codelist` in `attribute.py`

Adds the ability to create new attribute values from a list or to be generated from the provided cube dataframe. The unit test for this checks that each newly created value is unique.


csvqb/csvqb/models/cube/csvqb/components/attribute.py
```diff
@@ -31,10 +30,27 @@ def validate_data(self, data: PandasDataTypes) -> List[ValidationError]:
...
 
 @dataclass
-class NewQbAttribute(QbAttribute, UriIdentifiable):
+class NewQbAttributeValue(UriIdentifiable):
     label: str
     description: Optional[str] = field(default=None, repr=False)
-    code_list: Optional[QbCodeList] = field(default=None, repr=False)
+    uri_safe_identifier_override: Optional[str] = field(default=None, repr=False)
+    source_uri: Optional[str] = field(default=None, repr=False)
+    parent_attribute_uri: Optional[str] = field(default=None, repr=False)
```
robons: I think this should be `parent_attribute_value_uri`. As it is, `parent_attribute_uri` might confuse people and result in them saying that this attribute value is a child of an attribute - which they shouldn't do.

csvqb/csvqb/models/cube/csvqb/components/attribute.py
```diff
@@ -20,10 +33,17 @@ class QbAttribute(ColumnarQbDataStructureDefinition, ABC):
...
 
+    @abstractmethod
+    def new_attribute_values(self) -> List[NewQbAttributeValue]:
+        pass
+
 
 @dataclass
 class ExistingQbAttribute(QbAttribute):
     attribute_uri: str
+    new_attribute_values: List[NewQbAttributeValue] = field(
```
robons, line 44: @CharlesRendle just an FYI that I decided that `ExistingQbAttribute`s should be able to define new_attribute_values as well.

robons: I'm not sure whether it's worth having a `from_data` method here though, since this isn't the easy path for a user. If they're doing something as exotic as creating `NewQbAttributeValue`s whilst using an `ExistingQbAttribute`, I think they can write their own code to figure out what those `NewQbAttributeValue`s are.

robons (commented):
I've pretty much taken over this PR since @CharlesRendle isn't working today and we need this done ASAP.

csvqb/csvqb/models/cube/csvqb/components/attribute.py
canwaf: Agree. Defining `NewQbAttributeValue` using an `ExistingQbAttribute` cannot be done with an airgapped computer, so surfacing this functionality means "Congrulations, your library's scope is now 400% larger."

csvqb/csvqb/models/cube/csvqb/components/attribute.py
```diff
@@ -8,10 +8,23 @@
...
 from .datastructuredefinition import ColumnarQbDataStructureDefinition
-from .codelist import QbCodeList, NewQbCodeList
 from csvqb.models.validationerror import ValidationError
-from csvqb.inputs import PandasDataTypes
-from csvqb.models.cube.csvqb.catalog import CatalogMetadata
+from csvqb.inputs import PandasDataTypes, pandas_input_to_columnar_str
+
+
+@dataclass
+class NewQbAttributeValue(UriIdentifiable):
```
canwaf, line 16: ~I think this misnamed, and it took me a while to figure out why. To me `value` is a reserved word in my headspace for objects in triples, not the definition of the predicates storing the values of a component. We need to talk about this before I approve this. As always I might need this explaining to me if I'm misunderstanding the RDF Qb spec.~

Is this basically an attribute codelist? If so, shouldn't the identifier be a uri_safe transformation of the label by default?

csvqb/csvqb/models/cube/csvqb/components/attribute.py
```diff
@@ -52,11 +74,16 @@ def from_data(
         source_uri: Optional[str] = None,
         is_required: bool = False,
         uri_safe_identifier_override: Optional[str] = None,
-    ):
+    ) -> "NewQbAttribute":
+        columnar_data = pandas_input_to_columnar_str(data)
+        new_attribute_values_from_column = [
+            NewQbAttributeValue(v) for v in sorted(set(columnar_data))
+        ]
```
canwaf, line 81: This is not going to be performant on larger datasets.

canwaf (approved):

canwaf (approved):

csvqb/csvqb/models/cube/csvqb/components/attribute.py
robons: Yes it is, and that's why the `get_identifier` method returns the label, it's used by the `uri_safe_identifier` property defined in `UriIdentifiable`

Issue #82, opened 2021-07-12
Test Existing QB Validation Errors


robons:
validation tests (do not use the info.json interface to set these tests up, use the internal common model). Use the utils.qb.cube.validate_qb_component_constraints(...) method to perform validation in these tests.

- [x] Define a cube with no dimensions. Ensure that validation yields a suitable warning.
- [x] Define a cube with a QbSingleMeasureObservationValue with a unit defined. Also define a QbMultiUnits column. Only one of these should be defined; ensure that validation yields a suitable warning.
- [x] Define a cube with a QbSingleMeasureObservationValue but do not provide a unit. Do not define a QbMultiUnits column. A unit should be provided in one of these locations; ensure that validation yields a suitable warning.
- [x] Define a cube with more than one QbMultiUnits columns. Only one should be defined; ensure that validation yields a suitable warning.
- [x] Define a cube with more than one QbObservationValue columns. Only one should be defined; ensure that validation yields a suitable warning.
- [x] Define a cube with a QbMultiMeasureObservationValue, but do not define a QbMultiMeasureDimension; ensure that validation yields a suitable warning.
- [x] Define a cube with a QbMultiMeasureObservationValue, and define more than one QbMultiMeasureDimension. Only one should be defined; ensure that validation yields a suitable warning.
- [x] Define a cube with a QbSingleMeasureObservationValue and also define a QbMultiMeasureDimension. These types are incompatible; ensure that validation yields a suitable warning.

Issue #80, opened 2021-07-08
Ensure attributes can point to literals


canwaf:
Attributes can have literal values; ensure that definitions of attributes support literal values.

Two new classes `QbExistingAttributeLiteral` and `QbNewAttributeLiteral`.

* Keeping in mind that literals must have a xml-defined type
* That a language for strings *should* have a language (e.g. `"This is a literal string."@en`)
* That the attribute has `rdfs:range`

This ticket includes determining the inheritance from the Attribute parent class or the non-literal Existing/New subclasses.

canwaf:
- [ ] Raise a separate `lang_tag` ticket since language needs to be assigned on a literal basis not on a column basis

Implementation is adding `RDFS.range` to `AttributeProperty()` as an optional propertie. This is surfaced to the end-user as a `data_type` property to the new `NewQbAttributeLiteral` class.

The `ExistingQbAttributeLiteral` class is essentially a pass method, but we need the user to provide the same `data_type` value (because we cannot guarantee we can navigate to the existing attribute's definition uri to get the `data_type` for the column. If the user wants multiple data types within a single column they should use another utility to generate csvw.

In all cases the value of data_type is optional and if not provided it is assumed to be `xs:string`.

PR #122, opened 2021-08-06
Issue #82 - Making validation errors more type-specific & adding some unit tests

robons:


csvqb/csvqb/models/validationerror.py
```diff
@@ -2,11 +2,19 @@
...
-        self.message: str = message
+    """Class representing an error validating a model."""
 
-    def __repr__(self) -> str:
-        return f"{self.__class__.__name__}({self.message})"
+    message: str
+
+
+@dataclass
+class SpecificValidationError(ValidationError, ABC):
```
canwaf, line 17: Speaking of nonsense?

robons: What?

csvqb/csvqb/tests/unit/cube/qb/test_cubeqb_errorvalidation.py
```diff
@@ -111,10 +120,329 @@ def test_existing_dimension_output_uri_template():
...
+    assert error.component_type == QbMultiUnits
+    assert error.maximum_number == 1
+    assert error.actual_number == 2
+
+
+def test_multiple_obs_val_columns():
+    """
+    Ensure that when a user defines multiple observation value columns, we get an error.
+
+    We only currently accept the `MeasureDimension` style of multi-measure datasets.
```
canwaf, line 271: :soon:

csvqb/csvqb/tests/unit/cube/qb/test_cubeqb_errorvalidation.py
```diff
@@ -111,10 +120,329 @@ def test_existing_dimension_output_uri_template():
...
+    assert isinstance(error, MaxNumComponentsExceededError)
+    assert error.component_type == QbMultiMeasureDimension
+    assert error.maximum_number == 1
+    assert error.actual_number == 2
+
+
+def test_measure_dimension_with_single_measure_obs_val():
+    """
+    Ensure that when a user defines a measure dimension with a single-measure observation value, they get an error.
+    """
```
canwaf, line 405: Why is this an error? It's verbose but is it wrong?

robons: I see where you're coming from, however, I think they've fundamentally misunderstood something if they're taking this approach so I want to push back on them doing this. 

Don't forget that all of these validation errors are strictly advisory and don't stop the user actually doing anything.

canwaf (approved):
I look forward to trying to use dataclasses and pylance in my rewrite of #80 now that we've refactored QbAttribute and added a bunch of new errors. It would be good to see a consolidated list of errors so that they can be easily summoned when writing unit tests. I haven't checked if this shows up in a useful way in our docs.

robons:
> I haven't checked if this shows up in a useful way in our docs.

I forgot to include titles in the files so it isn't that clear where they are... but there's in there. They appear as if from nowhere. I'll fix and recommit.
image

PR #124, opened 2021-08-09
Robons 29

robons:


csvqb/csvqb/tests/unit/writers/test_skoscodelistwriter.py
```diff
@@ -0,0 +1,92 @@
...
+            "First Concept",
+            code="1st-concept",
+            description="This is the first concept.",
+        ),
+        NewQbConcept("Second Concept", parent_code="1st-concept", sort_order=20),
+    ],
+)
+
+
+def test_code_list_data_mapping():
```
canwaf, line 28: Got confused then realised this is a renamed file. I was about to ask you to stay in scope.

robons: Well, it's a bit more than that... the test file hadn't been migrated over to the pytest style... and I'm not 100% sure it was being run before.

canwaf (approved):

PR #128, opened 2021-08-16
Issue#79 Serialise Unit and Attribute Values alongside qb:DSD in CSV-W

santhosh-thangavel:
1. Serialized attribute values using serialise_attribute_values function.
2. Serialized unit using serialise_unit function.
3. Added test_serialise_new_attribute_values and test_serialise_unit functions.
4. Added test cases to see if new attribute values and unit are serialised.


csvqb/csvqb/tests/unit/writers/test_qbwriter.py
```diff
@@ -597,5 +596,205 @@ def test_about_url_generation_with_multiple_measures():
...
+            assert new_attribute_value.uri_str == expected_config["uri"]
+            assert new_attribute_value.comment == expected_config["description"]
+            assert (expected_config["source_uri"] is None and new_attribute_value.source_uri is None) \
+                   or str(new_attribute_value.source_uri.uri) == expected_config["source_uri"]
+            assert (expected_config[
+                        "parent_unit_uri"] is None and new_attribute_value.parent_attribute_value_uri is None) \
+                   or str(new_attribute_value.parent_attribute_value_uri.uri) == expected_config[
+                       "parent_unit_uri"]
+        else:
+            assert new_attribute_value is None
```
robons, line 794: Why don't we just remove the items from the dictionary @canwaf if we don't want to test them here?

csvqb/csvqb/writers/qbwriter.py
```diff
@@ -54,13 +57,16 @@ def write(self, output_folder: Path):
         # todo: Local units haven't been defined anywhere yet!
         # todo: Unit multiplier functionality hasn't been output.
 
+        see_also = rdf_resource_to_json_ld(self._generate_qb_dataset_dsd_definitions())
+        for x in self._serialise_attribute_values():
+            see_also += rdf_resource_to_json_ld(x)
+
```
robons, line 63: You need to make sure you serialise the units here too!

csvqb/csvqb/tests/unit/writers/test_qbwriter.py
santhosh-thangavel: Removed items from the dictionary which are not used for testing in test_serialise_unit function.

csvqb/csvqb/writers/qbwriter.py
santhosh-thangavel: unit serialization is added

robons (approved):
I've made some small changes that you may want to take a look at @santhosh-thangavel. 

@CharlesRendle - it would be a good idea if you could take a look at where we went with the code for when you're back.

Issue #38, opened 2021-06-25
Confirm CSV-W output would convert to valid RDF


canwaf:
Users should be able to confirm whether the resulting RDF would be valid according to the SPARQL tests

Try to elminate SPARQL failures when loading in dataframes/specifying metadata without having to run SPARQL queries (e.g., throw an error if you try to specify a dimension which has blanks)

We'll need one feature to test single measure datasets and another one to test multi-measure datasets.

PR #132, opened 2021-08-18
Issue #38 - Ensuring single + multi-measure cubes pass QB/SKOS SPARQL test constraints.

robons:


devtools/devtools/behave/temporarydirectory.py
```diff
@@ -14,8 +18,20 @@ def get_context_temp_dir_path(context) -> Path:
 
         context.add_cleanup(lambda: context.temp_dir.cleanup())
 
-        context.add_cleanup(
-            lambda: shutil.move(context.temp_dir.name, "/root/last_run_behave")
-        )
+        if "KEEP_FILES" in os.environ and bool(distutils.util.strtobool(os.environ["KEEP_FILES"])):
```
robons, line 21: @canwaf - I've hidden your functionality for keeping the files behind here and tidied it up a bit so you don't end up with all those `tmp` folders building up in there.

csvqb/csvqb/models/cube/qb/catalog.py
```diff
@@ -32,7 +33,7 @@ def get_description(self) -> Optional[str]:
         return self.description
 
     def get_identifier(self) -> str:
-        return self.title
+        return self.identifier or self.title
```
santhosh-thangavel, line 36: why we need an identifier?

csvqb/csvqb/tests/behaviour/qbwriter.feature
```diff
@@ -1,15 +1,15 @@
 Feature: Test outputting CSV-Ws with Qb flavouring.
 
   Scenario: A QbCube should generate appropriate DCAT Metadata
-    Given a single-measure QbCube named "Some Qube"
+    Given a single-measure QbCube with identifier "qb-id-10002" named "Some Qube"
```
santhosh-thangavel, line 4: If an identifier is not given, would we have a QbCube with a title?

robons: Yes, `title` is a required property. And if the user doesn't provide it, then they're ignoring all of our advice.

csvqb/csvqb/models/cube/qb/catalog.py
robons: So the identifier just allows us to differentiate between datasets on the SPARQL end-point without having to use the title. And it gives the user an extra degree of flexibility to specify an identifier that's better/more specific than the title.

PR #146, opened 2021-08-20
Robons 71

robons:


rossbowen:
I don't see that we need to choose between OM2 and QUDT, it seems reasonable we could make use of both.

For units like `MillionGBP` I'd still like to specify a `qudt:converstionFactor`. It feels inconvenient to _have_ to go via the base unit. (Though `MillionGBP` is one we'd end up defining centrally anyway).

How would a user specify something like "megalitres per day"?

robons:
> I don't see that we need to choose between OM2 and QUDT, it seems reasonable we could make use of both.
> 
I'm not entirely sure what you're suggesting here... I think we are using both. 

> For units like `MillionGBP` I'd still like to specify a `qudt:converstionFactor`. It feels inconvenient to _have_ to go via the base unit. (Though `MillionGBP` is one we'd end up defining centrally anyway).

I think if we take this approach we might want to take a more organised and systematic approach to the QUDT ontology and ensure we define all the related `quantityKinds`, etc. that we think we should be doing. It's probably something we would want to do more thoroughly than the quick-and-dirty approach that we're taking here to make it easy for users to do *something* as expressive as possible quite quickly.

> 
> How would a user specify something like "megalitres per day"?

i think they'd base it off the `m^3 per day` qudt unit (http://qudt.org/vocab/unit/M3-PER-DAY) and provide the conversion factor of `10^3`.

`1 megalitre = 10^6 litres = 10^6 * 10^-3 m^3 = 10^3 m^3`

csvqb/csvqb/models/cube/qb/components/unit.py
```diff
@@ -24,28 +25,45 @@
...
+    """The unit that this new unit is based on."""
+    base_unit_scaling_factor: Optional[float] = field(default=None, repr=False)
+    """
+    How to scale the value associated with this unit to map back to the base unit.
+    
+    e.g. if the base unit is *meters* and this unit (*kilometers*) has a scaling factor of **1,000**, then you multiply 
+    the value in *kilometers* by **1,000** to get the value in *meters*.    
+    """
+
+    optional_attribute_dependencies = enforce_optional_attribute_dependencies(
```
santhosh-thangavel, line 61: Please, Can you explain what the line of code from 61 to 66 does?

csvqb/csvqb/models/rdf/newunitresource.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+
+    source_uri: Annotated[
+        Optional[ExistingResource],
+        Triple(RDFS.isDefinedBy, PropertyStatus.recommended, map_resource_to_uri),
+    ]
+
+    base_unit_uri: Annotated[
+        MaybeResource["NewUnitResource"],
+        Triple(OM2.hasUnit, PropertyStatus.optional, map_resource_to_uri),
```
CharlesRendle, line 34: `NewUnitResource` extending `NewMetadataResouce` - We could't understand what the triple class is doing and why we hadn't used it on other classes which extend `NewMetadataResource`

csvqb/csvqb/models/cube/qb/components/unit.py
```diff
@@ -24,28 +25,45 @@
...
     label: str
     description: Optional[str] = field(default=None, repr=False)
-    unit_multiplier: Optional[int] = field(default=None, repr=False)
-    parent_unit_uri: Optional[str] = field(default=None, repr=False)
     source_uri: Optional[str] = field(default=None, repr=False)
     uri_safe_identifier_override: Optional[str] = field(default=None, repr=False)
     arbitrary_rdf: List[TripleFragmentBase] = field(default_factory=list, repr=False)
+    base_unit: Optional[QbUnit] = field(default=None, repr=False)
+    """The unit that this new unit is based on."""
+    base_unit_scaling_factor: Optional[float] = field(default=None, repr=False)
```
CharlesRendle, line 53: Should the default `base_unit_scaling_factor` be default to `1` instead of `None`? What difference would this have?

csvqb/csvqb/models/cube/qb/components/unit.py
CharlesRendle, line 61: We didn't understand what was going on here.
The notation - dictionary with key/value pair `str/list of str`
Why are attribute dependencies related to unit scaling factors?

sharedmodels/sharedmodels/tests/unit/rdf/test_resource_serialisation.py
```diff
@@ -42,18 +42,25 @@ class B(A):
 
     graph = b.to_graph(Graph())
 
-    assert 1 == len(graph)
+    assert len(graph) == 2
```
CharlesRendle, line 45: What has caused the graph length to change?

CharlesRendle (commented):
@santhosh-thangavel and I have worked through the changes made by this PR and have come up with some questions regarding the gaps in our combined knowledge.

Issue #125, opened 2021-08-11
Replace data strings with uri-safe-values


robons:
When the user decides to make use of the `from_data` functions to generate **units**/**measures**/**attribute values**/**code list values** from existing data labels, we take the existing string/label value from the cell and convert it into a *uri-safe format* which becomes part of the RDF resource's URI. We can't currently generate valid CSV-qb files using the `from_data` methods because we're not altering the user's data to point at the new *uri-safe* values; for example all of the code-list values wouldn't match up with the `notation` values defined in code lists. 

We need to write a function (possibly in the utils.qb package) which takes a `cube` and manipulates the data so that all string/label values are replaced with their *uri-safe* counterparts.  

We of course should have unit tests to ensure that the mapping occurs correctly.

PR #155, opened 2021-08-26
Issue #125 - Replacing labels with uri-safe-values where possible

robons:
Upon serialising a QbCube to CSV-W, we now replace all of the known labels with their appropriate local URI-safe values. This ensures that when a user defines a `NewQbDimension.from_data(...)`, we map their data to the new URI-safe values which define the codes. Similar mappings are written for `NewQbAttributeValue`s, as well as `NewQbUnit`s inside units columns and `NewQbMeasure`s inside measure dimension columns.

Satisfies issue #125


csvqb/csvqb/tests/unit/utils/qb/test_standardise.py
```diff
@@ -0,0 +1,171 @@
...
+        "Some Dimension": {"a", "b", "c"},
+        "Some Resource Attribute": {"r-a", "r-b", "r-c"},
+        "Measure": {"meas-a", "meas-b", "meas-c"},
+        "Unit": {"unit-a", "unit-b", "unit-c"},
+    }
+    expected_non_categorical_columns = {"Observed Value", "Some Literal Attribute"}
+
+    for (column_name, expected_categories) in map_col_to_expected_categories.items():
+        values = cube.data[column_name].values
+        assert isinstance(values, Categorical)
```
santhosh-thangavel, line 65: Is there a chance for the values to be an empty string or Null value? If so what would be the outcome?

robons: It shouldn't happen, but if it ever did, it'd cause the unit-test to fail which I think is perfectly fine - we'll know something is wrong.

csvqb/csvqb/tests/unit/utils/qb/test_standardise.py
```diff
@@ -0,0 +1,171 @@
...
+    ensure_qbcube_data_is_categorical,
+    convert_data_values_to_uri_safe_values,
+)
+
+
+def test_qbcube_data_is_categorical():
+    """
+    Ensure that when we convert a QbCube's dataframe columns to categorical, it is done for the appropriate columns.
+    """
+    data = pd.DataFrame(
```
santhosh-thangavel, line 16: Why do we convert QbCube’s DataFrame columns(appropriate columns) to Categorical data type?

robons: Because it allows us to do a more performant replacement of the column's values.

santhosh-thangavel:
@robons Thanks

santhosh-thangavel (approved):

PR #166, opened 2021-09-08
Fixing a couple of bugs I noticed.

robons:
- Ensures that you can create attribute values from a column containing missing values (attributes are not required by default). 
- Ensure that the observation's columns in the CSV-W point to the correct propertyUrl for the measure dimension/constant measure
- Ensure that the csvqb `.whl` package actually contains some code.


robons:
Jenkins build failure => closing and re-opening.

PR #168, opened 2021-09-08
Robons bug fix

robons:
- Ensures that you can create attribute values from a column containing missing values (attributes are not required by default).
- Ensure that the observation's columns in the CSV-W point to the correct propertyUrl for the measure dimension/constant measure
- Ensure that the csvqb .whl package actually contains some code.

### DIFF
--- a/csvqb/csvqb/inputs.py
+++ b/csvqb/csvqb/inputs.py
@@ -10,10 +10,10 @@ PandasDataTypes = Union[pd.DataFrame, pd.Series, None]
 
 
 def pandas_input_to_columnar(
-    maybe_columnar_data: PandasDataTypes, allow_none: bool = True
+    maybe_columnar_data: PandasDataTypes, allow_no_data_at_all: bool = True
 ) -> Optional[pd.Series]:
     if maybe_columnar_data is None:
-        if allow_none:
+        if allow_no_data_at_all:
             return None
         else:
             raise Exception("Columnar data not provided.")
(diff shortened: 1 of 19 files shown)


santhosh-thangavel (approved):

csvqb/csvqb/inputs.py
```diff
@@ -28,20 +28,39 @@ def pandas_input_to_columnar(
...
 
-def pandas_input_to_columnar_str(
-    maybe_columnar_data: PandasDataTypes, allow_none: bool = True
-) -> Iterable[str]:
-    series = pandas_input_to_columnar(maybe_columnar_data, allow_none)
+def pandas_input_to_columnar_optional_str(
+    maybe_columnar_data: PandasDataTypes,
+    allow_no_data_at_all: bool = True,
+) -> Iterable[Optional[str]]:
+    series = pandas_input_to_columnar(maybe_columnar_data, allow_no_data_at_all)
```
canwaf, line 44: I know where this is being used but what is the actual usecase? We don't allow sparse columns.

devtools/devtools/behave/csv2rdf.py
```diff
@@ -15,11 +15,13 @@
 from devtools.helpers.tar import dir_to_tar, extract_tar
 from devtools.behave.temporarydirectory import get_context_temp_dir_path
 
+client = docker.from_env()
+client.images.pull("gsscogs/csv2rdf")
```
canwaf, line 19: 🙌

canwaf (approved):
All sensible changes. Approved.

csvqb/csvqb/inputs.py
robons: Just confirming, we do allow sparse attributes.

PR #204, opened 2021-10-20
Update to toml to now surface scripts

canwaf:
Script section is required, @robons, please approve soon.


canwaf:
> Nooooooooo, for development purposes it needs to be kept as `./../sharedmodels`. 

Can't, because poetry cannot reference subdirectories in git, nor can it access outside its directory scope when building.

PR #221, opened 2021-10-29
Upgrade to rdflib 6.0.*

### LINKED ISSUE #177: Upgrade to rdflib 6.0.*
We're currently using rdflib 5.0.0 due to the fact that there are some breaking changes holding us back from using 6.0.*. 

When this happens, we should remove the use of rdflib-jsonld from anywhere that uses it (including csvqb & pmd). [rdflib itself how has functionality for (de-)serialising JSON-LD](https://github.com/RDFLib/rdflib-jsonld/tree/0.6.2#archived). 

Investigate the types of breaking changes implemented, the things we need to fix, and how to fix each type of thing.

Make sure that the solutions are portable so that `gss-utils` can adopt this approach to migrating from `rdflib~=5`.

Create sub-issues for this epic following the initial spike.

canwaf:
* Complete, only one change required (DCTerms updated to have title by rdflib)
Closes #177


robons:
> Looks like there's a bug in Pyright on imports
> 
> ```
> /workspaces/csvwlib/csvcubed-models/csvcubedmodels/rdf/dcat.py
>   /workspaces/csvwlib/csvcubed-models/csvcubedmodels/rdf/dcat.py:5:37 - error: "DCAT" is not exported from module "rdflib.namespace"
>     Import from "rdflib.namespace._DCAT" instead (reportPrivateImportUsage)
> ```
> 
> It can be fixed by changing the imports to `from rdflib.namespace._FOAF import FOAF` but that is inideal.
> 
> And then it becomes whackamol. Not quite there on everything else, new issues showing up.



> Looks like there's a bug in Pyright on imports
> 
> ```
> /workspaces/csvwlib/csvcubed-models/csvcubedmodels/rdf/dcat.py
>   /workspaces/csvwlib/csvcubed-models/csvcubedmodels/rdf/dcat.py:5:37 - error: "DCAT" is not exported from module "rdflib.namespace"
>     Import from "rdflib.namespace._DCAT" instead (reportPrivateImportUsage)
> ```

Seems odd we can't import those standard prefixes/namespaces anymore, the documentation suggests it's possible - https://rdflib.readthedocs.io/en/stable/apidocs/rdflib.namespace.html#importable-namespaces

> It can be fixed by changing the imports to `from rdflib.namespace._FOAF import FOAF` but that is inideal.
> 
> And then it becomes whackamol. Not quite there on everything else, new issues showing up.

I'll leave this PR be then, enjoy your whack-a-mole.

canwaf:
I'll take another look then.

Issue #151, opened 2021-08-25
Write PMD Conversion & Upload Pipeline - Part 1


robons:
- [x] Augment DCAT metadata with PMD-specific stuff (see issue #105)
- ~Run the date/time codelist generation tool.~ (Date-time codelist generation moved to infojson2csvqb build command).
- [x] Run the functionality to replace all CSV-relative URIs with ones relative to the base_uri.
  - to be decided in Issue #61 exactly what the base_uri will look like.
  - to be decided whether we write this in python to make it eaiser to unit test or whether we just use a bash command
 like `sed` and write some testing using the Jenkins test harness tests we have in pmd-jenkins-library.
  - see #213 for the implementation of this.
- [x] run csvlint
  - [x] ensure it fails when validation fails (integration test)
- [x] run csv2rdf
  - [x] ensure build fails when this process fails (integration test)

PR #222, opened 2021-11-02
Issue #151 - Adding CLI functionality to pull a CSV-W

robons:
and all of its relative depdendencies to a local filesystem folder.

This functionality is necessary to import CSV-Ws from a given URL as part of the PMD pipeline.

As part of this task, I've re-worked the existing CLI in pmdutils to use the [click](https://click.palletsprojects.com/en/8.0.x/) library.


csvcubed-devtools/csvcubeddevtools/helpers/file.py
```diff
@@ -0,0 +1,17 @@
+from pathlib import Path
+
+
+def get_test_cases_dir(start_dir: Path = Path(".")):
```
robons, line 4: I've moved this into its own folder so it can more easily (and sensibly) be referenced by other projects.

PR #241, opened 2021-11-15
Release_pydantic_relocation

canwaf:
Now using csvcubed-pydantic, same as v0.1.0-rc1.

### DIFF
--- a/csvcubed-devtools/pyproject.toml
+++ b/csvcubed-devtools/pyproject.toml
@@ -7,7 +7,7 @@ name = "csvcubed-devtools"
 packages = [
   {include = "csvcubeddevtools"},
 ]
-version = "0.0.1"
+version = "0.1.0-rc1"
 
 [tool.poetry.dependencies]
 behave = "*"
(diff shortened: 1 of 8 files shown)


csvcubed-devtools/pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "csvcubed-devtools"
 packages = [
   {include = "csvcubeddevtools"},
 ]
-version = "0.0.1"
+version = "0.1.0-rc1"
```
robons, line 10: I wouldn't do this. The version number is to be updated by the auto-tag-version-thingy as described in the release proceedure.

robons: Same for each of the other version changes you've made too.

canwaf: I didn't make any changes, I created the branch from v0.1.0-rc1. The new thing will overwrite them anyways.

robons: Ok, but you're merging it into main so it's confusing who sets/where the version gets set.

robons (approved):

PR #253, opened 2021-11-29
Upgrading rdflib to 6.0.x

robons:


canwaf (approved):
Does what it says on the tin. No more tombstoned versions. Yay!

PR #262, opened 2021-12-08
Muazzam chaud/issue211

MuazzamChaud:


csvcubed/csvcubed/utils/validators/uri.py
```diff
@@ -10,7 +10,7 @@
 from ..uri import ensure_looks_like_uri
 
 
-def validate_uri(attr_name: str, is_optional: bool = False) -> classmethod:
+def validate_uri(attr_name: list[str], is_optional: bool = False) -> classmethod:
```
robons, line 13: Instead of changing this functionality which is likely used elsewhere, it would be a better idea to add a new validator function and leave the existing one as-is.

csvcubed/csvcubed/utils/validators/uri.py
```diff
@@ -23,6 +23,9 @@ class SomeModel(PydanticModel):
 
             _some_uri_validator = validate_uri("some_uri")
     """
-    return validator(attr_name, allow_reuse=True, always=not is_optional)(
-        ensure_looks_like_uri
-    )
+    # attr_name = list(attr_name)
+    # attr_name = list(attr_name.split(" "))
+    for name in attr_name:
+        return validator(name, allow_reuse=True, always=not is_optional)(
```
robons, line 29: This won't work as expected. The function `validate_uri` function should always take a single `attr_name` since it is the name of the attribute to be inspected inside the model. It returns a validator which can be asked to validate the contents of `model.attr_name` at a later point in time.

csvcubed/csvcubed/models/cube/qb/catalog.py
```diff
@@ -21,7 +21,7 @@ class CatalogMetadata(CatalogMetadataBase, UriIdentifiable):
     description: Optional[str] = field(default=None, repr=False)
     creator_uri: Optional[str] = field(default=None, repr=False)
     publisher_uri: Optional[str] = field(default=None, repr=False)
-    landing_page_uri: Optional[str] = field(default=None, repr=False)
+    landing_page_uri: Union[str, list[str], None] = field(default_factory=list, repr=False) # todo: fix type on left hand side
```
robons, line 24: I'd be tempted to rename this to `landing_page_uris` and leave the type as just `list[str]`. We don't need to support multiple different shapes here, a list will cover providing no information `[]`, a single value `['http://some-uri']` or multiple values `['http://first-uri', 'http://second-uri']`.

csvcubed-models/csvcubedmodels/rdf/dcat.py
```diff
@@ -63,8 +63,8 @@ class Resource(NewMetadataResource):
         Triple(DCAT.keyword, PropertyStatus.recommended, map_str_to_en_literal),
     ]
     landing_page: Ann[
-        Optional[str], Triple(DCAT.landingPage, PropertyStatus.recommended, URIRef)
-    ]
+        Set[list[str]], Triple(DCAT.landingPage, PropertyStatus.recommended, URIRef)
+    ]  # todo: Enable support for multiple landing pages here, using a Set.
```
robons, line 67: Make sure you initialise the set in the `__init__` function. See what we're doing with themes/keywords there.

csvcubed/csvcubed/models/cube/qb/catalog.py
```diff
@@ -70,7 +72,7 @@ def configure_dcat_dataset(self, dataset: dcat.Dataset) -> None:
         dataset.license = self.license_uri
         dataset.creator = self.creator_uri
         dataset.publisher = self.publisher_uri
-        dataset.landing_page = self.landing_page_uri ########################################
+        dataset.landing_page = self.landing_page_uris ########################################
```
robons, line 75: I'd be tempted to add them individually instead of overwriting something that's already there.

PR #266, opened 2021-12-13
Muazzam chaud/issue211part2

MuazzamChaud:


csvcubed/csvcubed/models/cube/qb/catalog.py
```diff
@@ -11,7 +11,7 @@
 
 from csvcubed.models.cube.catalog import CatalogMetadataBase
 from csvcubed.models.uriidentifiable import UriIdentifiable
-from csvcubed.utils.validators.uri import validate_uri, validate_uris
+from csvcubed.utils.validators.uri import validate_uri, validate_uris_in_lists
```
robons, line 14: I would have gone for the singular `list` and the end here, but ok.

csvcubed/csvcubed/utils/uri.py
```diff
@@ -79,9 +79,9 @@ def ensure_looks_like_uri(value: str) -> None:
         raise ValueError(f"'{value}' does not look like a URI.")
         
 
-def ensure_looks_like_uris(values: list[str]) -> None:
+def ensure_values_in_lists_looks_like_uris(values: list[str]) -> None:
     """
-    Ensure that :obj:`value` looks like a URI.
+    Ensure that :obj:`value` in a list looks like a URI.
```
robons, line 84: So here the `value` we're referring to was the name of the function's parameter which you've renamed to `values`. So something like ```Ensure that the values in a list (:obj:`values`) look like URIs.``` would work here.

robons (commented):
Looks good, but just a couple of comments.

robons (approved):

Issue #271, opened 2021-12-17
`Suppressed` Values fail CSV-W Validation Check


robons:
Many of the cubes of open data which are produced for Government are sparesly populated (i.e. we don't have values for every possible combination of dimensions). Often this is because the data isn't relevant to the anticipated use-cases, but sometimes it's because the data has been actively suppressed by the publisher.

When the publisher actively suppresses an observation, we represent this by setting an `sdmxa:obsStatus` against the observation to explain why the value is not present. In this case, we say it's okay not to supply an observed value - since any value we input could accidentally influence an aggregated query a user could run over the dataset. However, we have still marked the column as `required` in the CSV-W schema; this means that although we're happy that the data is sound, the CSV-W doesn't pass validation.

We need to:

- [x] Allow users to avoid specifying a observation value where an `sdmxa:obsStatus` is set against the observation.
- [x] Continue to enforce the `required` nature of the CSV-W column where no `sdmxa:obsStatus` column has been set.
- [x] Apply validation in csvcubed to ensure each row either has an observed value or the cube has an attribute extending (or uses directly) the `sdmxa:obsStatus` attribute AND the observation has a value for said `sdmxa:obsStatus`.

PR #272, opened 2021-12-17
Issue #271 - Allow missing values when an `sdmxa:obsStatus` Attribute explains why it's missing.

robons:
Issues raised as a result of this PR:

* #273
* #274

Satisfies Issue #271.


csvcubed/csvcubed/utils/qb/cube.py
```diff
@@ -151,9 +164,73 @@ def _validate_observation_value_constraints(cube: Cube) -> List[ValidationError]
             errors += _validate_observation_value(obs_val_column, multi_units_columns)
             errors += _validate_single_measure_cube(cube, obs_val_column)
 
+        errors += _validate_missing_observation_values(cube, observed_value_columns[0])
+
     return errors
 
 
+def _validate_missing_observation_values(
```
canwaf, line 172: I think that the observations should be in their own module.

robons: I think at some point we'll end up splitting out all of the different validations into separate files. What makes you suggest it now?

csvcubed/tests/behaviour/qbwriter.feature
```diff
@@ -427,4 +427,26 @@ Feature: Test outputting CSV-Ws with Qb flavouring.
...
+    When the cube is serialised to CSV-W
+    Then csvlint validation of "good-qube.csv-metadata.json" should succeed
+
+  Scenario: Observation Values are Required where an `sdmxa:ObsStatus` Attribute Column is present but no value is set.
+    Given a single-measure QbCube named "Bad Qube" with missing observation values and missing `sdmxa:obsStatus` replacements
+    Then the CSVqb should fail validation with "Missing value(s) found for 'Value' in row(s) 0"
+    When the cube is serialised to CSV-W
+    # Unfortunately, CSV-W validation will *not* catch this error since the obs column cannot be marked as `required`
+    # since an `sdmxa:obsStatus` Attribute column has been defined.
+    Then csvlint validation of "bad-qube.csv-metadata.json" should succeed
```
canwaf, line 451: Is there a thing to do with csvcheck that could solve this?

csvcubed/csvcubed/utils/qb/cube.py
canwaf: It's grown this module by 50% LoC, which is some artibtrary mental threshold for some minor refactoring. Won't hurt if you don't want to do it now, but it is easily done now.

canwaf (approved):
I think pulling the observation stuff out of the cube.py is worthwhile now; however not a dealbreaker.

csvcubed/tests/behaviour/qbwriter.feature
robons: I don't think csvw-check is the right place to solve the issue. We'd effectively be making an unofficial extension to their CSV-W spec.

However, there is definitely scope for more CSV-qb validation which I believe is in our issues long-term.

PR #277, opened 2021-12-23
Muazzam chaud/issue202 Set up static docs

MuazzamChaud:


external-docs/mkdocs-static-site/docs/about.md
```diff
@@ -0,0 +1,38 @@
+# Qui Tres divitior
```
robons, line 1: Let's not commit any of the files in `mkdocs-static-site` since they can be auto-generated on Jenkins.

Jenkinsfile
```diff
@@ -136,6 +136,14 @@ pipeline {
                 }
             }
         }
+        stage('Build Static Site'){    
+            steps{
+                script{
+                    stash name: 'mkdocs', includes: '**/mkdocs-static-site/**/*'
```
robons, line 142: The mkdocs build step should be in here too so we build on commit.

external-docs/mkdocs-static-site/docs/about.md
robons: Well, except for the markdown files of course! But can you move them out of the `mkdocs-static-site` directory into `external-docs`? There's no point having a directory which only contains a directory.

robons (approved):
Looks like it all successfully builds and works (when Jenkins doesn't get in the way).

Could you add a new ticket into zen/github to set up a github repo to host these static sites (the API docs + this mkdocs stuff) and alter the Jenkins job to publish to the appropriate repo? It feels like a separate issue and you've covered this one off.

Make sure you update the title/description here though.

Issue #187, opened 2021-09-29
Tech Docs on qb Classes & How to use them


robons:
- The class structure for representing columns/qb-components.

So we need to show:

* how these classes can be used to build a cube
  * using the from_data methods
* how you can add additional RDF to individual components
* performing validation

robons:
Task completed. Documentation on how to use these classes can be found [here](https://github.com/GSS-Cogs/csvcubed/blob/main/csvcubed/csvcubed/models/cube/qb/README.md#defining-a-qbcube).

PR #282, opened 2021-12-29
Issue #187 - Adding documentation on using Cube/QB Classes.

robons:
The latest formatted version of the documentation can be found [here](https://github.com/GSS-Cogs/csvcubed/blob/robons-187/csvcubed/csvcubed/models/cube/qb/README.md#defining-a-qbcube).

I've moved a number of the QB components around into a modular structure I'm a bit happier with.


osamede20 (approved):

osamede20 (approved):

canwaf (approved):
If there is ever a case for a `! Danger zone` format in Markdown, it's this entire document.

Great primer on how to access the QbCube's low-level API which generally we don't want people to use.

I added some module docstrings I found missing, and edited the main documentation to avoid cliches and for brevity.

Issue #254, opened 2021-11-30
Set up Documentation to Guide Users through Error Handling / Debugging (Troubleshooting)


robons:
Set up some documentation which provides a lookup from common error messages to guidance on what they mean and how they can generally be solved. 

- [x] We should have one page per error type. (<20 different error types).
  - [x] Put some limited description of what the error means and generic resolution steps (in plain english).
- [ ] Set up some short-urls for these documents so we don't fill the terminal up with too much useless information that makes the user's eyes glaze over. (todo: We need to figure out whether which URL shortners we're allowed to use - I think there's a gov.uk URL shortner somewhere?)
  - We should not be using common third-party URL shortners; see [here](https://gcs.civilservice.gov.uk/blog/link-shorteners-the-long-and-short-of-why-you-shouldnt-use-them/) for an explaination. Speak to @robons to make use of the http://purl.org/csvcubed domain he's registered as an initial solution for this.

- [x] Set up a github issue OR confluence page where our internal users can provide reports of stack traces / validation errors when they happen so we can keep track of the frequency at which different errors occur for users.

robons:
I've added a google sheet to help keep track of validation errors people find when using csvcubed. It can be found [here](https://docs.google.com/spreadsheets/d/1Gyx4bRCxIquVA6gb5mQeaf0x_wc5gLZNo0uUvfLUSP8/edit?usp=sharing).

PR #288, opened 2022-01-07
Issue #254 - Adding documentation for external users to help understand and fix errors.

robons:
Reworked some of the exceptions to make them more specific - to help make it easier for users to understand what a given error message means.

Done as part of #254.


canwaf (approved):
Cool!

Issue #287, opened 2022-01-06
Re-work csvcubed Resource URIs


robons:
- [x] Shorten concept scheme URIs. They should be `#{notation}` and `#codelist` instead of `#concept/{scheme-name}/{notation}` and `#scheme/{scheme-name}` respectively. The catalogue metadata URIs should probably continue to contain their type in their name since the concepts are the primary identifiers of interest in the CSV-W.
  - [x] Add a validation check which ensures that none of the concepts in the scheme happen to have the uri-identifier of `codelist` so that a concept called `codelist` cannot clash with the scheme definition.
- [x] Alter **all** relative URLs so that they never use a slash `/` after the fragment identifier (unless there is a parent/child relationship between the two parts of the URL). We need to replace all these instances of `/` with another character. Use a semicolon (`;`) for now -- but make sure this parameter comes from a configurable value in the `WriterBase` class so we can easily alter the functionality in future.
- [x] Add the capability to alter how we deal with defining URIs for each file. We currently use `./{data-file-name}.csv#` to differentiate resources declared in different files. Add some functionality to easily customise this.

**N.B. this task will require reworking a *lot* or behave tests in both csvcubed and gss-utils**.

robons:
Not using semi-colon `;` since it's a [reserved character](https://www.w3.org/Addressing/URL/uri-spec.html#z8)

robons:
Once this has been published, @canwaf agrees to act as spokesperson to present this to the real world to see what they think about our URIs. 

Speak to the Greek Lord of the Qb + Swirrl.

PR #290, opened 2022-01-10
Issue #287 - Altering how URIs defined in CSV-Ws.

robons:
PR to alter how URIs are configured in serialised CSV-Ws.

All URI fragments use the permitted chars as defined in the [w3c URI spec](https://www.w3.org/Addressing/URL/uri-spec.html#z8)

Current URIs look like this:

* Data Structure Definition:
  * `cube-name.csv#obs/{dimension_value1},{dimension_value2}@{measure_identifier}`
  * `cube-name.csv#component/{component_identifier}`
  * `cube-name.csv#measure/{measure_identifier}`
  * `cube-name.csv#unit/{unit_identifier}`
  * `cube-name.csv#attribute/{attribute_identifier}`
  * `cube-name.csv#attribute/{attribute_identifier}/{attribute_value_identifier}`
  * `cube-name.csv#dimension/{dimension_identifier}`
  * `cube-name.csv#class/{class_identifier}`
  * `cube-name.csv#dataset`
  * `cube-name.csv#structure`
* Code-list definition:
  * `code-list-name.csv#code-list` (scheme)
  * `code-list-name.csv#{concept-identifier}`


csvcubed/csvcubed/writers/qbwriter.py
```diff
@@ -330,7 +325,7 @@ def _get_qb_units_column_specification(
         self, column_name_uri_safe: str
     ) -> rdf.qb.AttributeComponentSpecification:
         component = rdf.qb.AttributeComponentSpecification(
-            self._doc_rel_uri(f"component/{column_name_uri_safe}")
+            self._new_uri_helper.get_component_uri(column_name_uri_safe)
```
robons, line 328: bug

rossbowen:
Ty for taking in feedback about the `/` awkwardness, though for now my initial reaction was `/` over `.`.

Also I really do prefer `codelist` to `code-list`. 😂

csvcubed/csvcubed/utils/pandas.py
```diff
@@ -122,3 +135,23 @@ def coalesce_on_uri_safe(data: PandasDataTypes) -> PandasDataTypes:
...
+        return is_string_dtype(data.cat.categories)
+
+    return is_string_dtype(data)
+
+
+def _get_unique_values(data: PandasDataTypes) -> Iterable[str]:
+    if data is None:
+        return []
+
+    if data.dtype == "category":
```
canwaf, line 154: This won't work because you can't get dtype from a pd.DataFrame, it will need to be coalesced to a Series first.

csvcubed/tests/unit/cube/qb/test_cubeqb_errorvalidation.py
```diff
@@ -698,5 +699,43 @@ def test_attribute_numeric_resources_validation():
...
+            ),
+        ],
+    )
+    errors = qube.validate() + validate_qb_component_constraints(qube)
+    assert_num_validation_errors(errors, 1)
+    error = errors[0]
+    assert isinstance(error, ReservedUriValueError)
+    assert error.csv_column_name == "New Dimension"
+    assert error.reserved_identifier == "code-list"
+    assert error.conflicting_values == ["Code List"]
```
canwaf, line 737: I think this means that if a user wanted a code list with "code list" in it they could create a label of "code list" provided the value was something like "cd list", right?

canwaf (changes requested):
The `qbwriter.py` changes must have taken you ages, but this is a great example of parameterisation, which will help us if we ever have a change of heart on code lists.

That said, there are some problems with the PandasDataTypes class in pandas.py which need fixing. (There's a comment on the problem.)

Also unwritten validation should use `pass` instead of returning empty lists.

csvcubed/csvcubed/models/cube/qb/components/codelist.py
```diff
@@ -125,19 +125,21 @@ def get_permitted_rdf_fragment_hints(self) -> Set[RdfSerialisationHint]:
     def get_default_node_serialisation_hint(self) -> RdfSerialisationHint:
         return RdfSerialisationHint.ConceptScheme
 
-    def validate_data(self, data: PandasDataTypes) -> list[Optional[ValidationError]]:
+    def validate_data(
+        self, data: PandasDataTypes, column_csv_title: str
+    ) -> list[ValidationError]:
```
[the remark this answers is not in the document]
robons: absolutely

csvcubed/csvcubed/models/cube/qb/components/datastructuredefinition.py
```diff
@@ -35,6 +35,7 @@ def validate_data(
         data: pd.Series,
         column_csvw_name: str,
         csv_column_uri_template: Optional[str],
+        column_csv_title: str,
```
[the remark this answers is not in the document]
robons: lol, this is an abstract method implemented in various other locations.

csvcubed/csvcubed/models/cube/qb/components/dimension.py
```diff
@@ -62,7 +62,11 @@ def get_default_node_serialisation_hint(self) -> RdfSerialisationHint:
     _range_uri_validator = validate_uri("range_uri")
 
     def validate_data(
-        self, data: pd.Series, column_csvw_name: str, csv_column_uri_template: str
+        self,
+        data: pd.Series,
+        column_csvw_name: str,
+        csv_column_uri_template: str,
+        column_csv_title: str,
```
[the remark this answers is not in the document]
robons: We need it to return something of the expected type. `pass` won't match the type signature.

csvcubed/csvcubed/utils/pandas.py
robons: noooooooo

csvcubed/tests/unit/cube/qb/test_cubeqb_errorvalidation.py
robons: My initial answer was that as long as you provide a `uri_safe_override` of something which isn't `code-list`, you should be fine. 

But then I noticed we aren't actually checking any of the `uri_safe_override` values when we're enforcing uniqueness. Since the (python) user can provide such overrides, it's actually *those* values, not just the `uri_safe(label)` we've been working with so far that we need to ensure are unique.

Ugh.

csvcubed/csvcubed/models/cube/qb/components/validationerrors.py
```diff
@@ -68,3 +68,48 @@ class UndefinedAttributeValueUrisError(UndefinedValuesError):
...
+
+
+@dataclass
+class ReservedUriValueError(SpecificValidationError):
+    """
+    An error which occurs when the user has defined a resource which would re-use a reserved URI value.
+    """
+
+    csv_column_name: str
+    conflicting_values: List[str]
```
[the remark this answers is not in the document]
robons: This is the list of values that the user has in their code list which conflict with reserved keywords.

https://github.com/GSS-Cogs/csvcubed/blob/6ab5ed39cc6eee63c9002affe735e00ef36a189d/csvcubed/csvcubed/utils/pandas.py#L91

csvcubed/tests/unit/cube/qb/test_cubeqb_errorvalidation.py
robons: I suspect we'll need to spin this problem out into a separate issue.

csvcubed/csvcubed/utils/pandas.py
robons: I've made the input param type more specific to avoid this problem.

csvcubed/tests/unit/cube/qb/test_cubeqb_errorvalidation.py
robons: https://github.com/GSS-Cogs/csvcubed/issues/309

canwaf: Yuk

canwaf (approved):

Issue #195, opened 2021-10-12
Ensure compatibility with python 3.10


robons:
Python 3.10 [has recently been released](https://pythoninsider.blogspot.com/2021/10/python-3100-is-available.html?utm_source=feedburner&utm_medium=feed&utm_campaign=Feed%3A+PythonInsider+%28Python+Insider%29) but causes some problems with our code. Unit tests are failing. We need to find out what's failing, how much of our code needs rewriting/fixing and whether that will impact on our ability to support python 3.9. 

To make it clear: we want to be able to support *both* Python 3.9 and 3.10, so let's not start using any 3.10-specific features if we can avoid it. We might need to do some if statements/branching depending on the version of python that the tool is being run under. e.g.

```python
if 3.9:
  do_thing()
elif 3.10:
  do_other_thing()
```

Ideally, we could set up multiple docker files for 3.9/3.10 so we can test that all tests pass in both containers before publishing. This doesn't necessarily have to be part of the standard Jenkins build, but could be part of a special build process we do when releasing. We're likely to want to extend this when 3.11, etc. are released in future.

If there is little work to do, do it as part of this task, else convert this into an epic with appropriate sub-tasks.

CharlesRendle:
TDL:

- [ ] Get BDD tests working through tox - think this is because the command needs to be executed from the .feature files location

- [ ] Unit tests are not working in the csvcubed package `ModuleNotFoundError: No module named 'csvcubedmodels'`

- [ ] Apply these changes to all subsequent packages (csvcubed-models, e.c.t)

- [ ] Write script to run the tox command for all packages (run x3)

- [ ] Github action to trigger this script

CharlesRendle:
Jenkins file/ tox stage now starts the container from `pythonversiontesting` directory successfully but the subsequent tox commands are not recognised. Jenkins returns `tox: not found`

robons:
@CharlesRendle: pydantic have just released v1.9.0 which has python 3.10 support - I've released an updated csvcubed-version of the project with our bug fixes on top. Upgrading the version of csvcubed-pydantic in the csvcubed project should stop the errors you were seeing coming from pydantic when using python 3.10.

https://pypi.org/project/csvcubed-pydantic/1.9.0/

PR #302, opened 2022-01-12
Charles rendle/issue195

CharlesRendle:
Ensure compatibility with python 3.10 #195

PR adds a step to the Jenkins build pipeline (Tox) which runs all of our tests once in the current python version and once again in an updated version of python. The python versions used in testing are controlled from a new Dockerfile in the pythonversiontesting directory. The jenkinsfile Test and Tox stages should be triggered by git tag conditionals. Each library has a Tox.ini file which runs the testing commands for pyright, pytest and behave.

Current Issues:

- Running pyright has been disabled in the csvcubed package as well as in the Tox.ini belonging to csvcubed because of a number of errors which need to be sorted before merging.

- The syntax of the git tag conditionals is most definitely wrong. Following: https://www.jenkins.io/blog/2018/05/16/pipelines-with-git-tags/ tag discovery is enabled but we want to run the Tox stage only when a release tag is being pushed and run the usual Test stage in all other events.


thomasmastersmethods:
Tested locally and no major issues with this, docker can handle any path issues - there are some pep8 issues across the project that would perhaps need to be addressed which is unrelated to this issue however

robons (approved):

PR #304, opened 2022-01-17
MuazzamChaud/issue268 Add Validation Error where Existing Measure Dimension is Missing Value URI Template

MuazzamChaud:


csvcubed/tests/unit/utils/qb/validation/test_observations.py
```diff
@@ -57,5 +62,59 @@ def test_find_sdmxa_obs_status_columns():
...
+            )
+        ],
+    )
+    errors = _validate_multi_measure_cube(qube, None)
+    assert len(errors) == 0, [e.message for e in errors]
+
+
+def test_value_uri_template_is_missing_in_existing_measure_dimention():
+    """
+    Testing to see that the value uri template is not defined when the measure dimention is defined.
```
robons, line 95: I'd alter this to make it clear we should get an error.

csvcubed/tests/unit/utils/qb/validation/test_observations.py
```diff
@@ -57,5 +62,59 @@ def test_find_sdmxa_obs_status_columns():
...
+            QbColumn(
+                "Measure", 
+                QbMultiMeasureDimension([
+                    ExistingQbMeasure("http://some-measure")
+                ]),
+            )
+        ],
+    )
+    errors = _validate_multi_measure_cube(qube, None)
+    assert len(errors) == 1, [e.message for e in errors]
```
robons, line 117: Can you add some assertions to ensure the error is of the expected type and points at the right type of component?

csvcubed/tests/unit/utils/qb/validation/test_observations.py
```diff
@@ -57,5 +62,59 @@ def test_find_sdmxa_obs_status_columns():
     }, obs_status_col_names
 
 
+def test_value_uri_template_is_present_in_existing_measure_dimention():
+    """
+    Testing to see that when the measure dimention is defined, that the value uri template is also defined.
```
robons, line 67: I'd say something like:

> Ensuring that there is no validation error raised for a multi-measure column which re-uses existing measures when a `csv_column_uri_template` is provided.

csvcubed/csvcubed/utils/qb/validation/observations.py
```diff
@@ -147,11 +151,28 @@ def _validate_multi_measure_cube(
...
+        measure_column: QbColumn[QbMultiMeasureDimension] = multi_measure_columns[0]
+
+        all_measures_existing = all(
+            [
+                isinstance(m, ExistingQbMeasure)
+                for m in measure_column.structural_definition.measures
+            ]
+        )
+    
+        if all_measures_existing:
```
robons, line 169: it would be neater to just have one if block `if all_measures_existing and measure_column.csv_column_uri_template is None:`

robons (commented):
Almost there, just test a bit more and clarify comments.

csvcubed/tests/unit/utils/qb/validation/test_observations.py
```diff
@@ -57,5 +63,63 @@ def test_find_sdmxa_obs_status_columns():
...
+                "Measure", 
+                QbMultiMeasureDimension([
+                    ExistingQbMeasure("http://some-measure")
+                ]),
+            )
+        ],
+    )
+    errors = _validate_multi_measure_cube(qube, None)
+    assert len(errors) == 1, [e.message for e in errors]
+    assert 'CsvColumnUriTemplateMissingError' in str(errors)
```
robons, line 121: It's safer to do something like the below:

```python
error = errors[0]
assert isinstance(error, CsvColumnUriTemplateMissingError0
assert error.component == ExistingQbMeasure
```

This way, if (and when) we refactor the names of these things, the tests should automatically get refactored too. And we're checking the right attributes match.

csvcubed/tests/unit/utils/qb/validation/test_observations.py
```diff
@@ -57,5 +63,64 @@ def test_find_sdmxa_obs_status_columns():
...
+        ],
+    )
+    errors = _validate_multi_measure_cube(qube, None)
+    assert len(errors) == 0, [e.message for e in errors]
+
+
+def test_value_uri_template_is_missing_in_existing_measure_dimention():
+    """
+    Ensuring that there is a validation error raised,
+    for a multi-measure column which re-uses existing measures when a csv_column_uri_template is provided.
```
robons, line 98: Should be
> is [**not**] provided.

robons (approved):
Just try and fix that comment before merging in!

PR #305, opened 2022-01-17
Draft qbconfig.jsonc file

canwaf:
This is the example qbconfig.jsonc file for review, which covers off the last of the comments from the info.json review session.

Further issues need to be created in the repo to cover off interpretation; however if this is good to go we have an example which can be used to start the schema design.


robons (approved):
Tip top

PR #308, opened 2022-01-21
MuazzamChaud/issue 256 : purl short urls for each validation error

MuazzamChaud:


csvcubed/tests/unit/cube/qb/test_cubeqb_errorvalidation.py
```diff
@@ -123,6 +123,9 @@ def test_existing_dimension_csv_column_uri_template():
     assert isinstance(validation_error, CsvColumnUriTemplateMissingError)
     assert validation_error.csv_column_name == "Existing Dimension"
 
+    assert isinstance(CsvColumnUriTemplateMissingError.get_error_url(), str)
+    assert CsvColumnUriTemplateMissingError.get_error_url() == 'http://purl.org/csv-cubed/err/csv-col-uri-temp-mis'
```
robons, line 127: It'd make me feel more comfortable to use `validation_error.get_error_url()` here since we need the error URL to be linked to the particular error instance.

robons: Same in all the other locations.

csvcubed/csvcubed/models/cube/validationerrors.py
```diff
@@ -4,6 +4,7 @@
 """
 from collections import Set
 from dataclasses import dataclass
+from typing import ClassVar
```
robons, line 7: Can we get rid of this now?

csvcubed/csvcubed/models/cube/qb/validationerrors.py
```diff
@@ -6,7 +6,7 @@
 """
 
 from dataclasses import dataclass
-from typing import Optional, Type, Union
+from typing import ClassVar, Optional, Type, Union
```
robons, line 9: Can we get rid of `ClassVar` now?

robons (commented):
Mostly looks good, but just ensure we can access the error URL from the object/instance of the class in the tests.

robons (approved):

Issue #285, opened 2022-01-04
List Dimensions/Attributes/Measures/Units to reuse


robons:

robons:
@rossbowen have you been able to take a look at this yet?

PR #312, opened 2022-01-24
Draft Cube Config Schema json.

Giraffe-Technology:
Appears to be a lack of differentiation between New and Existing column types.
Column attributes could be extracted into lower level $defs.

### DIFF
--- a/csvcubed/csvcubed/schema/cube-config-schema.jsonc
+++ b/csvcubed/csvcubed/schema/cube-config-schema.jsonc
@@ -381,7 +381,8 @@
             },
             "from_existing": {
               "description": "The URI to an existing set of values",
-              "$ref": "http://gss-cogs.github.io/family-schemas/external.json#/definitions/attr_ext_uri"
+              "type": "string",
+              "format": "uri"
             },
             "definition_uri": {
               "description": "The URI link to an existing definition",
--- a/csvcubed/csvcubed/schema/cube-config-schema.jsonc
+++ b/csvcubed/csvcubed/schema/cube-config-schema.jsonc
@@ -23,14 +23,14 @@
             "description": "URI that defines the spatial / geographic bounds of the data contained herein",
             "type": "string",
             "format": "uri",
-            "const": "https://tdg.docbook.org/tdg/publishers/5.1/dcterms.html#dcterms.spatial"
+            "default": "https://tdg.docbook.org/tdg/publishers/5.1/dcterms.html#dcterms.spatial"
         },
 
         "temporal_bound": {
             "description": "URI that defines the temporal bounds of the data contained herein",
             "type": "string",
             "format": "uri",
-            "const": "https://tdg.docbook.org/tdg/publishers/5.1/dcterms.html#dcterms.temporal"
+            "default": "https://tdg.docbook.org/tdg/publishers/5.1/dcterms.html#dcterms.temporal"
         },
 
         "columns": {
@@ -61,34 +61,34 @@
 			"description": "Definition of units for the observed value",
 			"type": "object",
 			"properties": {
-				"anyOf": {
-					"label": {
-						"description": "Replacement label for the Unit",
-						"type": "string"
-					},
-					"from_existing": {
-						"description": "URI of an Existing Qb Dimension",
-						"type": "string",
-						"format": "uri"
-					},
-					"scaling_factor": {
-						"type": "decimal"
+				"label": {
+					"description": "Replacement label for the Unit",
+					"type": "string"
+				},
+				"from_existing": {
+					"description": "URI of an Existing Qb Dimension",
+					"type": "string",
+					"format": "uri"
+				},
+				"scaling_factor": {
+					"description": "The scaling factor",
+					"type": "decimal"
 
-					},
-					"si_scaling_factor": {
-						"type": "decimal"
+				},
+				"si_scaling_factor": {
+					"description": "The SI scaling factor",
+					"type": "decimal"
 
-					},
-					"quantity_kind": {
-						"description": "Source URI for the definition of the Unit Kind",
-						"type": "string",
-						"format": "uri"
-					},
-					"definition_uri": {
-						"description": "Source URI for the definition of the Unit",
-						"type": "string",
-						"format": "uri"
-					}
+				},
+				"quantity_kind": {
+					"description": "Source URI for the definition of the Unit Kind",
+					"type": "string",
+					"format": "uri"
+				},
+				"definition_uri": {
+					"description": "Source URI for the definition of the Unit",
+					"type": "string",
+					"format": "uri"
 				}
 			}
 		},
@@ -138,7 +138,7 @@
         "newDimension": {
             "description": "A new dimension",
             "$ref": "#/$defs/dimension",
-            "additionalProperties": {
+            "properties": {
 				"codelist": {
                     "description": "Source URI for the code-list for the dimension, e.g. 'http://example.org/code-list/{+some_column}'",
                     "type": "string",
@@ -152,7 +152,8 @@
             "type": "object",
             "properties": {
                 "type": {
-                    "type": "string",
+                    "description": "Column type, defaults to 'dimension'.",
+					"type": "string",
                     "default": "dimension"
                 },
                 "label": {
@@ -160,7 +161,7 @@
                     "type": "string"
                 },
                 "description": {
-                    "description": "",
+                    "description": "Description for the column",
                     "type": "string"
                 },
                 "from_existing": {
@@ -202,11 +203,15 @@
 				"type": {
                     "description": "The type of column",
                     "type": "string",
-                    "const": "attribute"
+                    "default": "attribute"
                 },
+				"description": {
+					"description": "A description of the attribute",
+					"type": "string"
+				},
 
                 "from_existing": {
-                    "description": "The type of column",
+                    "description": "The URI of an existing attribute column",
                     "type": "string",
                     "format": "uri"
                 },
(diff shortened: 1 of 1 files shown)


csvcubed/csvcubed/schema/cube-config-schema.jsonc
```diff
@@ -0,0 +1,430 @@
...
+        },
+
+        "temporal_bound": {
+            "description": "URI that defines the temporal bounds of the data contained herein",
+            "type": "string",
+            "format": "uri",
+            "const": "https://tdg.docbook.org/tdg/publishers/5.1/dcterms.html#dcterms.temporal"
+        },
+
+        "columns": {
```
robons, line 36: `columns` should be a dictionary of column name to the object type.

csvcubed/csvcubed/schema/cube-config-schema.jsonc
```diff
@@ -0,0 +1,303 @@
...
+            "description": "URI that defines the temporal bounds of the data contained herein",
+            "type": "string",
+            "format": "uri",
+            "default": "https://tdg.docbook.org/tdg/publishers/5.1/dcterms.html#dcterms.temporal"
+        },
+
+        "columns": {
+            "description": "Columns in the same order as the csv, unspecified column types are assumed to be dimensions, case insensitive",
+			"properties": {
+				"Existing dimension": {"$ref": "#/$defs/existingDimension"},
```
robons, line 39: These are the example column names that @canwaf used in his JSON. The columns in CSVs won't necessarily have these names so we need to be able to support *any* column name mapping to any column type.

csvcubed/csvcubed/schema/cube-config-schema.jsonc
```diff
@@ -0,0 +1,303 @@
...
+		"unit": {
+			"description": "Definition of units for the observed value",
+			"type": "object",
+			"properties": {
+				"label": {
+					"description": "Replacement label for the Unit",
+					"type": "string"
+				},
+				"from_existing": {
+					"description": "URI of an Existing Qb Dimension",
```
robons, line 69: The `from_existing` properties should provide users with a selection of pre-defined values they can select from as well as allowing them to override that selection with their own string value. 

See [here](https://github.com/GSS-Cogs/family-schemas/blob/main/dataset-schema-1.1.0.json#L820) for an example of how that works in our existing schemas. This definition enables vscode to provide a list of suitable values to select from.

The same behaviour should happen for all other `from_existing` properties, but they should be pointing at the relevant URIs for that *type*.

csvcubed/csvcubed/schema/cube-config-schema.jsonc
```diff
@@ -0,0 +1,303 @@
...
+		"existingDimension": {
+            "description": "An existing dimension",
+            "$ref": "#/$defs/dimension"
+        },
+
+        "newDimension": {
+            "description": "A new dimension",
+            "$ref": "#/$defs/dimension",
+            "properties": {
+				"codelist": {
```
robons, line 142: @canwaf I think we need to correct this clear spelling error ;)

csvcubed/csvcubed/schema/cube-config-schema.jsonc
```diff
@@ -0,0 +1,303 @@
...
+					"items": {"$ref": "#/$defs/attributeValue"}
+				}
+			}
+        },
+
+		"attributeValue": {
+			"description": "A single attribute value, optionally referencing existing values and definition URIs",
+			"type": "object",
+			"properties": {
+				"label": {
```
robons, line 249: `label` should be required in attribute values/units/measures.

csvcubed/csvcubed/schema/cube-config-schema.jsonc
```diff
@@ -0,0 +1,303 @@
...
+					"type": "string",
+					"description": "A label for the attribute values"
+				},
+				"description": {
+					"type": "string",
+					"description": "The description of the attribute values"
+				},
+				"existing_value": {
```
robons, line 257: @canwaf should `existing_value` follow the `from_existing` pattern we're using elsewhere?

csvcubed/csvcubed/schema/cube-config-schema.jsonc
robons: Also, both `new` and `existing` dimension should both have the codelist property available on them.

csvcubed/csvcubed/schema/cube-config-schema.jsonc
canwaf: Yes.

csvcubed/csvcubed/schema/cube-config-schema.jsonc
canwaf: Yes, I've talked with @Giraffe-Technology and we've got a better handle on this one now.

csvcubed/csvcubed/schema/cube-config-schema.jsonc
canwaf: There are some exceptions, though. We won't let anyone name a column "observations" and have it be a dimension. Additionally, there are some names we don't want (e.g. `code-list`, its variations, and other common RDF vocabulary used to construct URIs)

robons: > We won't let anyone name a column "observations" and have it be a dimension
I'm not sure we should be *that* prescriptive. If that's what they want to call a dimension, who are we to disagree?

csvcubed/csvcubed/schema/cube-config-schema.jsonc
canwaf: Could you explain this one a little further. We're not currently proposing to fetch the existing values from the defined component? My understanding is that would have to be done when we load the qube_config into csvcubed?

robons: I want this: 

image

robons: Admittedly, I suspect our list will be a lot shorter, and depends on @rossbowen 's work in #285

csvcubed/csvcubed/schema/cube-config-schema.jsonc
canwaf: Are we joking about "code list" being made "uri-safe" by omitting the space instead of using a dash? Do we need to update our style guide?

csvcubed/csvcubed/schema/cube-config-schema.jsonc
robons: > The contents of a column

When I define a list of units against a multi-units column, how do you know which unit definition matches up with which label?

When i define a unit against an observation column, where on earth does the unit's label come from?

csvcubed/csvcubed/schema/cube-config-schema.jsonc
canwaf: Okay. "Here's this rope, do what you will."

csvcubed/csvcubed/schema/cube-config-schema.jsonc
canwaf: Okay. I'm fine with this. `code-list` it is.

csvcubed/csvcubed/schema/cube-config-schema.jsonc
Giraffe-Technology: Resolved, the column name can be anything (string), on adding the 'type' property an enum of the available 'columns' is available.  This offers dimension, attribute, measure, unit and observation.
Note: these values are dynamically generated from the type: const: string values.

csvcubed/csvcubed/schema/cube-config-schema.jsonc
Giraffe-Technology: I've updated the 'from_existing' attributes to refer to ($ref") the family schemas http://gss-cogs.github.io/family-schemas/external.json#/definitions/...

csvcubed/csvcubed/schema/cube-config-schema.jsonc
Giraffe-Technology: Have updated codelist to code-list, without clear schema differences there is only 'dimension' and not 'new dimension' / 'existing dimension'.

csvcubed/csvcubed/schema/cube-config-schema.jsonc
```diff
@@ -1,303 +1,515 @@
...
+              "description": "The URI that defines the attribute source",
+              "anyOf": [
+                {
+                  "description": "The relative file path to the code-list's local definition",
+                  "type": "string",
+                  "format": "uri"
+                },
+                {
+                  "description": "The URI of the codelist associated with the dimension.",
+                  "$ref": "http://gss-cogs.github.io/family-schemas/external.json#/definitions/attr_ext_uri"
```
robons, line 169: Definition URIs link to things like PDF documents a user wants to publish and shouldn't provide any auto-complete help like you've got here. These URIs aren't suitable since they're not helpful to a human.

csvcubed/csvcubed/schema/cube-config-schema.jsonc
```diff
@@ -1,303 +1,515 @@
...
+              "type": "string",
+              "description": "A label for the attribute values"
+            },
+            "description": {
+              "type": "string",
+              "description": "The description of the attribute values"
+            },
+            "from_existing": {
+              "description": "The URI to an existing set of values",
+              "$ref": "http://gss-cogs.github.io/family-schemas/external.json#/definitions/attr_ext_uri"
```
robons, line 421: This isn't the right help to be providing to the user here. Attributes and attribute values are not the same kind of thing. Attribute values are something an attribute has whereas you're linking to the attributes themselves here. Leave it without intellisense/help for now.

Giraffe-Technology: updated.

csvcubed/csvcubed/schema/cube-config-schema.jsonc
Giraffe-Technology: updated.

canwaf (approved):
This now passes my tests (observation column with single unit/measure defined, and measure/unit columns). Great job, everything connects as it should now and the keywords are all how I envisioned.

There are a few items which I would consider a bit rough around the edges (e.g. accepting false as a value in the `values` keys for units and measures which will instruct csvcubed to create new units based on the values within that column); however that kind of thing requires more thought and this isn't the last of this file. We'll find more things we can improve while we build out the csvcubed functionality to read in this file.

I'm merging it in to get it done and off the backlog.

PR #313, opened 2022-01-25
Initial logging configuration functionality.

robons:
Adding some bland colourless logging.

### DIFF
--- a/csvcubed/csvcubed/utils/log.py
+++ b/csvcubed/csvcubed/utils/log.py
@@ -1,0 +1,40 @@
+"""
+Log
+---
+
+Utilities to help with logging.
+"""
+import logging
+import sys
+from typing import Union
+from colorama import Fore, Style
+
+
+def start_logging(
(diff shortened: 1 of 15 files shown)


csvcubed/csvcubed/utils/log.py
```diff
@@ -0,0 +1,39 @@
+"""
+Log
+---
+
+Utilities to help with logging.
+"""
+import logging
+import sys
+from colorama import Fore, Style
```
canwaf, line 9: I don't think this meets a threshold for inclusion as a dependency in our package. I suggest removing it.

robons: I mean if we can do the same without it then fine we can remove it, but I'm inclined to include it until we come across a problem requiring us to remove it.

canwaf (approved):
Fixing things not germane to the ticket `is_optional=True`

Approved under the assumption that colorizer is removed

PR #321, opened 2022-02-02
Task 299

Giraffe-Technology:
Schema Version downgrade and rename from jsonc to json.


robons:
Functionality already covered in PR #332.

Issue #320, opened 2022-02-01
PMDifcation of dcat metatadata results in duplicate catalogue entries


robons:
Need to ensure that we have different catalogue entry URIs for each CSV-W even if the document's base URIs are only distinguished by a fragment identifier.

See [attached files](https://app.zenhub.com/files/374709947/51a283e4-fbd4-47b2-8f34-0be7bfe35179/download) or [here](https://ci.ukstats.dev/job/GSS_data/job/Trade/job/csvcubed/job/ONS-International-Trade-in-Services/job/ONS-International-Trade-in-Services/) for example.

robons:
![image.png](https://images.zenhubusercontent.com/60d35dd0922789587fc8922d/87740914-5dcd-4775-88ae-9f3b91089caf)

The build finally passes all of the SPARQL tests!

PR #330, opened 2022-02-10
Keep skos:ConceptScheme datasets

canwaf:
Resolves SPARQL failures caused by @robons #320 fixes.


canwaf:
No longer required, as this has been fixed correctly instead of SPARQL-based whack-a-mole.

PR #332, opened 2022-02-11
Draft Pull for Rt task 297

Giraffe-Technology:
Unit Tests are basic - still working on them


csvcubed/csvcubed/models/rdf/namespaces.py
```diff
@@ -0,0 +1,51 @@
+"""
+Namespaces
+----------
```
canwaf, line 3: These Namespaces should be imported form csvcubed-models.rdf.namespaces (the modules library exists to reference external models)

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,421 @@
+"""
+Models
+------
+
+info.json V1.1 column mapping models.
```
canwaf, line 5: This isn't info.json anymore.

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,572 @@
...
+    maybe_property_value_url: Optional[str],
+) -> Optional[QbCodeList]:
+
+    is_date_time_code_list = (
+        (
+            maybe_code_list is None
+            or (isinstance(maybe_code_list, bool) and maybe_code_list)
+        )
+        and maybe_parent_uri
+        == "http://purl.org/linked-data/sdmx/2009/dimension#refPeriod"
```
canwaf, line 89: Cool! Above and beyond the ticket.

csvcubed/csvcubed/models/cube/qb/catalog.py
```diff
@@ -29,11 +29,17 @@ class CatalogMetadata(CatalogMetadataBase, UriIdentifiable):
...
     public_contact_point_uri: Optional[str] = field(default=None, repr=False)
     uri_safe_identifier_override: Optional[str] = field(default=None, repr=False)
+    spatial_bound_uri: Optional[str] = field(default=None, repr=False)
+    temporal_bound_uri: Optional[str] = field(default=None, repr=False)
 
     _creator_uri_validator = validate_uri("creator_uri", is_optional=True)
     _publisher_uri_validator = validate_uri("publisher_uri", is_optional=True)
     _landing_page_uris_validator = validate_uris_in_list("landing_page_uris", is_optional=True)
     _license_uri_validator = validate_uri("license_uri", is_optional=True)
+    _spatial_bound_uri_validator = validate_uri("spatial_bound_uri", is_optional=True)
```
robons, line 39: These spatial and temporal bounds are technically a new feature which will require other changes in the underlying csvcubed models. @canwaf do we want to drop these for now and pick them up in a separate issue?

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,572 @@
...
+        except Exception as err:
+            log.error(f"{type(err)} exception raised because: {repr(err)}")
+
+    cube_metadata = _metadata_from_dict(
+        {"id": str(csv_path).upper(), "title": csv_path.name.upper().replace("_", " ")}
+    )
+    # TODO - Create the new Cube
+    cube = Cube(metadata=cube_metadata, data=data, columns=cube_columns)
+    # validation_errors = cube.validate()
+    validation_errors = cube.pydantic_validation()
```
robons, line 467: There's a second level of validations in [csvcubed.utils.qb.validation.cube.validate_qb_component_constraints](https://github.com/GSS-Cogs/csvcubed/blob/main/csvcubed/csvcubed/utils/qb/validation/cube.py#L21) that does some more in-depth validation of the cube's defintion - including checking for things like no dimensions/observations columns/etc.

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,572 @@
...
+
+    return cube, validation_errors
+
+
+def get_cube_from_info_json(
+        config_path: Path,
+        data: pd.DataFrame,
+        cube_id: Optional[str] = None,
+        log: logging.RootLogger = None
```
robons, line 476: Do we need to accept a log as input here since it's already defined at the top of the module?

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,572 @@
...
+    """
+    Generates a Cube structure from a config.json input.
+    :return: tuple of cube and json schema errors (if any)
+    """
+
+    config = _load_resource(config_path)
+    schema_path = Path("..", "..", "..", "..", "csvcubed", "schema", "cube-config-schema.json").resolve()
+    schema = _load_resource(schema_path)
+    try:
+        errors = validate_dict_against_schema(value=config, schema=schema, log=log)
```
robons, line 487: I think I prefer the approach where if there are schema validation errors, we present them to the user but allow the process to complete as best it can.

There are often schema validation errors which occur because someone is using a dimension URI which hasn't been defined in our list of known dimensions; the dimension URI may still be valid but we just haven't included it yet in our JSON schema.

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,572 @@
...
+    # csv_path: Path,
+    # fail_when_validation_error_occurs: bool,
+    # validation_errors_file_out: Optional[Path],
+):
+    # print(f"{Style.DIM}CSV: {csv_path.absolute()}")
+    # print(f"{Style.DIM}info.json: {info_json.absolute()}")
+    data: pd.DataFrame = pd.read_csv(csv_path)
+
+
+    if info_json:
```
robons, line 517: We need this to allow conventional column definitions at the same time as a user providing a `config.json`.

csvcubed/csvcubed/readers/configdeserialiser.py
robons: I see that the function which calls this runs that function, so I wouldn't return any validation errors here since they'll only be duplicates.

csvcubed/csvcubed/readers/mapcolumntocomponent.py
```diff
@@ -0,0 +1,165 @@
...
+    column_without_type = copy.deepcopy(column)
+    del column_without_type["type"]
+
+    if column_type is None:
+        raise ValueError("Type of column not specified.")
+    elif column_type == "dimension":
+        if any([column_without_type.get(a) for a in ['label', 'description', 'cell_uri_template', 'code_list']]):
+            # probably a New dimension, so move all properties under a new node
+            # 'new' can be bool or NewDimensionProperty (only former works)
+            column = {'type': column_type,
```
robons, line 114: Instead of doing this reshaping, the `schema.ExistingDimension` and `schema.NewDimension` classes should be rewritten so that they exactly match the properties available in the JSON document. 

They're designed to be classes which exactly match the style of JSON the user is expected to write.

robons: Let's have a call about this.

csvcubed/csvcubed/utils/file.py
```diff
@@ -33,3 +35,28 @@ def _ensure_dir_structure_exists(dir_path: Path) -> None:
         _ensure_dir_structure_exists(dir_path.parent)
         _logger.debug("Creating directory %s", dir_path)
         dir_path.mkdir()
+
+
+def read_json_from_file(file_path: Path, log: logging.RootLogger) -> dict:
```
robons, line 40: Again, I think the log here should be a log created specifically for this module instead of being accepted from the caller.

csvcubed/csvcubed/utils/uri.py
```diff
@@ -107,3 +110,34 @@ def ensure_values_in_lists_looks_like_uris(values: list[str]) -> None:
             raise ValueError(f"'{value}' does not look like a URI.")
 
     _logger.debug("Values %s all look like URIs.", values)
+
+
+def read_from_uri(uri: str, log: logging.RootLogger) -> dict:
```
robons, line 115: This feels more like a `load_json_from_uri` method. I think it would be better placed with the `file.read_json_from_file` method in a separate module called `json`.

robons: Again, I'm not sure why you're accepting a `log` as input param here.

csvcubed/csvcubed/utils/validators/schema.py
```diff
@@ -0,0 +1,36 @@
...
+        errors = jsonschema.validate(value, schema)
+
+    except jsonschema.exceptions.ValidationError as err:
+        log.error(f"Validation of the supplied config cube failed: {repr(err)}")
+
+    except jsonschema.exceptions.SchemaError as err:
+        log.error(f"The schema provided for validation of the config cube was not a valid schema: {repr(err)}")
+
+    except Exception as err:
+        print(repr(err))
```
robons, line 34: Shouldn't we be sticking this in the log too?

csvcubed/csvcubed/utils/validators/schema.py
```diff
@@ -0,0 +1,36 @@
+import logging
+
+import jsonschema
+from jsonschema.exceptions import ValidationError
+
+
+def validate_dict_against_schema(value: dict, schema: dict, log: logging.RootLogger) -> list[ValidationError]:
```
robons, line 7: Again, I think the logger should be owned by this module, not passed in.

robons (commented):
I've left a number of comments. I think we need to have a chat about the `schema.X` models and how we can rewrite those to match the JSON coming from the user's input.

csvcubed/csvcubed/models/rdf/namespaces.py
robons: I don't think this is resolved. Have a chat with @canwaf about this one.

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+T = TypeVar("T", bound=object)
+
+
+@dataclass
+class SchemaBaseClass(DataClassBase):
+    ...
+
+
+@dataclass
+class Dimension(SchemaBaseClass):
```
robons, line 53: Even if we have one underlying model in the JSON schema, can we not split this into two models in python? One for Existing and another for New?

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+        # label, description or code_list present therefore it will be a NewDimension
+        else:
+            new_dimension = NewQbDimension.from_data(
+                label=label,
+                data=data,
+                description=self.description,
+                parent_dimension_uri=self.from_existing,
+                source_uri=self.definition_uri,
+            )
+            # new_dimension.code_list = self._get_code_list(new_dimension, json_parent_dir)
```
robons, line 90: We need this to be uncommented.

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+    #         json_parent_dir: Path,
+    # ) -> Union[NewQbCodeListInCsvW, ExistingQbCodeList]:
+    #
+    #     code_list_obj = None
+    #
+    #     if isinstance(self.code_list, str):
+    #         if looks_like_uri(self.code_list):
+    #             code_list_obj = ExistingQbCodeList(self.code_list)
+    #         else:
+    #             code_list_path = Path(self.code_list)
```
robons, line 106: This part of the `else` statement should just yield an exception in csvcubed. We don't need to use `NewQbCodeListInCsvW` in this project.

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+    #         ):
+    #             # This is a special case where we build up a code-list of the date/time values.
+    #             code_list_obj = (
+    #                 self._get_date_time_code_list_for_dimension(self.label, new_dimension)
+    #             )
+    #         # else, the user wants a standard codelist to be automatically generated
+    #     else:
+    #         raise ValueError(f"Unmatched code_list value {self.code_list}")
+
+    # def _get_date_time_code_list_for_dimension(
```
robons, line 131: We definitely want this.

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+    definition_uri: Optional[str] = None
+
+
+@dataclass
+class NewAttributeProperty(SchemaBaseClass):
+    path: Optional[str] = None
+    label: Optional[str] = None
+    comment: Optional[str] = None
+    isDefinedBy: Optional[str] = None
+    subPropertyOf: Optional[str] = None
```
robons, line 165: What's this `subPropertyOf` stuff doing in here? That's from the old info.json v1.1 syntax.

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+    from_existing: Optional[str] = None
+    definition_uri: Optional[str] = None
+    data_type: Optional[str] = None
+    required: Optional[bool] = None
+    values: Union[bool, List[AttributeValue]] = True
+
+    def map_to_qb_attribute(
+        self, label: str, data: PandasDataTypes
+    ) -> Union[NewQbAttribute, ExistingQbAttribute]:
+        pass
```
robons, line 183: Why do we have two copies of the same function here?

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+                source_uri=self.definition_uri,
+                is_required=self.isRequired
+            )
+        else:
+            if isinstance(self.values, bool) and self.values is True:
+                return NewQbAttribute.from_data(
+                    label=column_title, data=data, is_required=self.required
+                )
+
+            elif isinstance(self, NewAttributeProperty):
```
robons, line 220: Will this ever be true?

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+    def map_to_existing_qb_attribute(
+        self, data: PandasDataTypes
+    ) -> Union[ExistingQbAttribute, ExistingQbAttributeLiteral]:
+        if self.literalValuesDataType is None:
+            return ExistingQbAttribute(
+                self.uri,
+                new_attribute_values=_get_new_attribute_values(
+                    data, self.newAttributeValues
+                ),
+                is_required=self.isRequired,
```
robons, line 243: `required`

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+                raise ValueError(f"Unhandled value: {self}")
+
+    def map_to_existing_qb_attribute(
+        self, data: PandasDataTypes
+    ) -> Union[ExistingQbAttribute, ExistingQbAttributeLiteral]:
+        if self.literalValuesDataType is None:
+            return ExistingQbAttribute(
+                self.uri,
+                new_attribute_values=_get_new_attribute_values(
+                    data, self.newAttributeValues
```
robons, line 241: `values`

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+
+            else:
+                raise ValueError(f"Unhandled measure type: {self.measure}")
+
+            return QbSingleMeasureObservationValue(
+                measure=measure, unit=unit, data_type=self.datatype or "decimal"
+            )
+
+
+def _map_unit(resource: Units) -> NewQbUnit:
```
robons, line 386: I don't think the input type is correct here. It should be `resource: UnitType` given your current naming.

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+            for unit in self.values:
+                if not isinstance(unit, UnitType):
+                    raise ValueError(f"Unexpected unit value: {unit}")
+
+                if any([unit.label, unit.description, unit.definition_uri]):
+                    # NewUnit
+                    units.append(_map_unit(unit))
+
+                elif unit.from_existing:
+                    units.append(ExistingQbUnit(unit_uri=self.values.from_existing))
```
robons, line 300: I don't think `self.values.from_existing` is going to resolve to anything since `self.values` is a list.

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+                else:
+                    raise ValueError(f"Unexpected unit value: {unit}")
+
+            return QbMultiUnits(units)
+
+        raise ValueError(f"Unhandled 'Units' value: {self}")
+
+
+@dataclass
+class MeasureType(SchemaBaseClass):
```
robons, line 311: Again, can we just call this `Measure` instead of `MeasureType`?

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+            new_measures = []
+            for new_measure in self.values:
+                if not isinstance(new_measure, MeasureType):
+                    raise ValueError(f"Unexpected measure: {new_measure}")
+
+                if any([new_measure.label, new_measure.description, new_measure.definition_uri]):
+                    # NewQBMeasure
+                    new_measures.append(_map_measure(new_measure))
+                else:
+                    new_measures.append(QbMultiMeasureDimension(new_measure))
```
robons, line 339: In this case it would be an `ExistingQbMeasure`, not a Multi-measure dimension.

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+            return QbMultiMeasureDimension(new_measures)
+
+        else:
+            raise ValueError(f"Unexpected values type: {self.values}")
+
+
+@dataclass
+class ObservationValue(SchemaBaseClass):
+    datatype: Optional[str] = None
+    unit: Union[None, str, Units] = None
```
robons, line 350: It should be `Union[None, str, UnitType]` and `Union[None, str, MeasureType]` using your current definitions. It's only possible to set the unit and measure on the observation when it's a single unit/measure for the whole dataset.

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+    measure: Union[None, str, Measures] = None
+
+    def map_to_qb_observation(self) -> QbObservationValue:
+        unit = None
+        if isinstance(self.unit, str):
+            unit = ExistingQbUnit(unit_uri=self.unit)
+
+        elif isinstance(self.unit, Units):
```
robons, line 358: Again, this needs to be `UnitType` here.

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,594 @@
...
+#         / f"{uri_safe(column_label)}.csv-metadata.json"
+#     )
+
+
+def initialise_from_config(self, config_json: dict) -> bool:
+    """
+    Initialises a Qube from a qube json file.
+
+    param: config_json: A Qube created according to a Qube Config schema
+    """
```
robons, line 135: This function doesn't seem to have an implementation yet. I know that @MuazzamChaud 's work has been merged into main now so it's probably merging his work into your branch and integrating with his CLI `build` command work.

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,594 @@
...
+    columns: List[QbColumn] = []
+    for column_title in d.get('columns', []):
+        column_config = d['columns'].get(column_title)
+        columns.append(map_column_to_qb_component(column_title, column_config,
+                                                  data[column_title], json_parent_dir))
+
+    return Cube(metadata, data, columns)
+
+
+def _override_config_for_cube_id(config: dict, cube_id: str) -> Optional[dict]:
```
robons, line 166: We should get rid of this now. It isn't necessary for this qube-config.json stuff.

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,594 @@
...
+    publisher = get_with_func_or_none(config, "publisher", lambda p: str(GOV[uri_safe(p)]))
+    themes = config.get('themes', "")
+    if isinstance(themes, str) and themes:
+        themes = [themes]
+    keywords = config.get('keywords', "")
+    if isinstance(keywords, str) and keywords:
+        keywords = [keywords]
+
+    return CatalogMetadata(
+        identifier=config.get("$id"),
```
robons, line 200: Do we really want to use the `$id` property? Surely `id` is good enough?

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,594 @@
...
+        title=get_from_dict_ensure_exists(config, "title"),
+        description=config.get("description", ""),
+        creator_uri=creator,
+        publisher_uri=publisher,
+        theme_uris=[uri_safe(t) for t in themes] if isinstance(themes, list) else None,
+        keywords=keywords if isinstance(keywords, list) else None,
+        spatial_bound_uri=uri_safe(config.get('spatial_bound', "")),
```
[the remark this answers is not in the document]
Giraffe-Technology: updated theme, spatial bound and temporal bound to avoid calling urisafe if a None / not present  /empty string

csvcubed/csvcubed/readers/configdeserialiser.py
Giraffe-Technology: $id seemed appropriate based on https://json-schema.org/draft/2020-12/json-schema-core.html#id-keyword
however I've updated to plain old id.

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+
+    # Properties only available for New Dimension
+    label: Optional[str] = None
+    description: Optional[str] = None
+    definition_uri: Optional[str] = None
+
+    # Properties common to both New and Existing Dimension
+    from_existing: Optional[str] = None
+    cell_uri_template: Optional[str] = None
+    code_list: Optional[Union[str, bool]] = False
```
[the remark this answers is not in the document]
Giraffe-Technology: Agreed. fixed.

csvcubed/csvcubed/readers/columnschema.py
Giraffe-Technology: This is contrary to previous guidance where you indicated that the schema classes should correspond to the json schema.
This is because there is no clear distinction between the object properties for NewDimension and ExistingDimension (etc.) 

For infojson the structure was clearly different with the 'new' property being present for New columns.
If this is really required then we need to discuss how to differentiate these.

Currently a call that is conditional on 'type' == 'dimension' to schema.**Dimension**.dict_fields_match_class confirms we have the correct fields before instantiating the Dimension dataclass, the branching to choose New/Existing is done in the map_to_qb_dimension method.

Feasibly we could set type to new_dimension or exsiting_dimension on the config json so the conditional logic containing the dict_fields_match_class could be updated to check the correct New/Existing dimension fields.

Alternatively the schema itself could be updated so that the properties contain either/and title or description to be considered a new dimension.  For other column types this needs checking.

csvcubed/csvcubed/models/cube/qb/catalog.py
```diff
@@ -72,6 +78,6 @@ def configure_dcat_dataset(self, dataset: dcat.Dataset) -> None:
         dataset.publisher = self.publisher_uri
         dataset.landing_page = set(self.landing_page_uris)
         dataset.themes = set(self.theme_uris)
-        dataset.keywords = set(self.keywords)
+        dataset.keywords = set(self.keyword)
```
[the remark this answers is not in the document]
Giraffe-Technology: sorry, missed that in update following your answer,  updated.

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,572 @@
...
+        "value",
+        "val",
+        "vals",
+    ],
+    "units": ["unit", "units", "units column", "unit column", "unit type", "unit types"]
+    # 'attribute': ['attribute', 'attributes']  # Not supported for loading cube from convention only.
+}
+
+log = logging.getLogger(__name__)
+start_logging()
```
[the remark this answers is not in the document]
Giraffe-Technology: removed calls to start_logging.

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,421 @@
...
+
+
+@dataclass
+class NewDimensionProperty(SchemaBaseClass):
+    codelist: Union[str, bool]
+    path: Optional[str] = None
+    label: Optional[str] = None
+    comment: Optional[str] = None
+    isDefinedBy: Optional[str] = None
+    subPropertyOf: Optional[str] = None
```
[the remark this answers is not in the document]
Giraffe-Technology: updated.

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,421 @@
...
+                description=self.new.comment,
+                parent_dimension_uri=self.new.subPropertyOf,
+                source_uri=self.new.isDefinedBy,
+                uri_safe_identifier_override=self.new.path,
+            )
+            if isinstance(self.new.codelist, str):
+                if looks_like_uri(self.new.codelist):
+                    new_dimension.code_list = ExistingQbCodeList(self.new.codelist)
+                else:
+                    code_list_path = Path(self.new.codelist)
```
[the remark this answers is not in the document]
Giraffe-Technology: I've updated the approach for getting the code_list from using the _get_code_list() method to NewQbCodeList.from_data(),
therefore this comment is outdated.

csvcubed/csvcubed/readers/configdeserialiser.py
Giraffe-Technology: Recent changes have removed this code as it belonged to the older infojson was of creating codelists

csvcubed/csvcubed/models/rdf/namespaces.py
Giraffe-Technology: the import had been updated, this was inadvertently left in, file is now deleted.

csvcubed/csvcubed/models/cube/qb/components/measuresdimension.py
```diff
@@ -34,8 +34,10 @@ class QbMultiMeasureDimension(QbColumnStructuralDefinition):
     @staticmethod
     def new_measures_from_data(data: PandasDataTypes) -> "QbMultiMeasureDimension":
         columnar_data = pandas_input_to_columnar_str(data)
+
+        qb_measures = [NewQbMeasure(m) for m in sorted(set(columnar_data))]
         return QbMultiMeasureDimension(
-            [NewQbMeasure(m) for m in sorted(set(columnar_data))]
+            qb_measures  # TODO investigate None value from generator
```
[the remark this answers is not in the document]
Giraffe-Technology: TODO removed, in adding an intermediate variable for debugging it resulted a None as the method being called is a generator method.

csvcubed/csvcubed/readers/configdeserialiser.py
Giraffe-Technology: Commented out, then deleted in later commit.

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,594 @@
...
+                column_name, QbMultiMeasureObservationValue(maybe_data_type)
+            )
+        else:
+            raise Exception(f"Unmatched column definition: {col_config}")
+
+    elif isinstance(col_config, bool) and col_config:
+        return SuppressedCsvColumn(column_name)
+    else:
+        # If not a known/expected type/value (or is a string), treat it as a dimension.
+        maybe_description: Optional[str] = None
```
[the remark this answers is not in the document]
Giraffe-Technology: this code is no longer used and has been removed.

csvcubed-pmd/csvcubedpmd/dcatcli/pmdify.py
```diff
@@ -374,7 +374,7 @@ def _get_catalog_entry_from_dcat_dataset(csvw_graph: Graph) -> pmdcat.Dataset:
             (GROUP_CONCAT(?landingPage ; separator='|') as ?landingPages) 
             (GROUP_CONCAT(?theme; separator='|') as ?themes) 
             (GROUP_CONCAT(?keyword; separator='|') as ?keywords) 
-            ?contactPoint ?identifier
+            ?contactPoint ?identifier ?datasetContents
```
robons, line 377: Why is this commit in the PR? It's already in master. I don't want duplicate commits if we can avoid it.

csvcubed/csvcubed/cli/build.py
```diff
@@ -1,13 +1,24 @@
...
+from csvcubedmodels.dataclassbase import DataClassBase
+
+from csvcubed.readers.configdeserialiser import (
+    get_cube_from_data,
+    get_cube_from_config_json
+)
+from csvcubed.utils.qb.validation.cube import validate_qb_component_constraints
+
 _logger = logging.getLogger(__name__)
+_logger.addHandler(logging.StreamHandler())
```
robons, line 21: We don't want to merge this one in.

Isn't the logger output configured in Muazzam's code? I know there were some bugs in it that I've fixed. If you merge a fresh copy of `origin/main` in it should make sure it works.

csvcubed/csvcubed/cli/build.py
```diff
@@ -19,5 +30,50 @@ def build(
...
+
+    else:
+        cube, json_schema_validation_errors = get_cube_from_data(csv_path)
+
+    validation_errors = cube.validate()
+    validation_errors += validate_qb_component_constraints(cube)
+
+    if not output_directory.exists():
+        print(f"Creating output directory {output_directory.absolute()}")
+        output_directory.mkdir()
```
robons, line 47: I think there's a param `parent=True` which creates any parent directories which don't exist. It's usually a good one to set so they can specify `out/some-other-folder` as the out dir and have it create both dirs.

csvcubed/csvcubed/cli/build.py
```diff
@@ -19,5 +30,50 @@ def build(
...
+
+    if len(validation_errors) > 0 or len(json_schema_validation_errors) > 0:
+        for error in validation_errors:
+            print(
```
robons, line 51: Can we have these output using `logger.error`? I want to make sure that info is in the log file.

csvcubed/csvcubed/cli/build.py
```diff
@@ -19,5 +30,50 @@ def build(
...
+                f"Validation Error: {error.message}"
+            )
+
+        for err in json_schema_validation_errors:
+            print(
```
robons, line 56: Can you output the schema validation errors as `logger.warning`?

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -184,197 +128,14 @@ def _metadata_from_dict(config: dict) -> "CatalogMetadata":
...
+def get_cube_from_data(csv_path: Path) -> QbCube:
     """
     Given a Pandas DataFrame, it resolves each column to a column based on naming conventions
     The cube must have dimension, observation, units and measures columns
     The csv_path arg is used for the Cube title
     """
+    data: pd.DataFrame = pd.read_csv(csv_path)
     if data.shape[0] < 2:
         # Must have 2 or more rows, a heading row and a data row
         raise ValueError("The csv data did not contain sufficient rows to continue")
```
robons, line 141: Let's change this message to something like `CSV input must contain header row and at least one row of data`.

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,572 @@
...
+    The cube must have dimension, observation, units and measures columns
+    The csv_path arg is used for the Cube title
+    """
+    if data.shape[0] < 2:
+        # Must have 2 or more rows, a heading row and a data row
+        raise ValueError("The csv data did not contain sufficient rows to continue")
+
+    elif data.shape[1] < 3:
+        # Must have 3 or more columns, 1+ dimension, 1 observation, 1 attribute,
+        raise ValueError("The csv did not contain sufficient columns to continue")
```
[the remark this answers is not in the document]
robons: Let's make sure to get rid of this.

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,331 @@
...
+) -> Tuple[QbCube, List[ValidationError]]:
+    """
+    Generates a Cube structure from a config.json input.
+    :return: tuple of cube and json schema errors (if any)
+    """
+    data: pd.DataFrame = pd.read_csv(csv_path)
+    config = _load_resource(config_path.resolve())
+    schema = _load_resource(Path(config['$schema']).resolve())
+    try:
+        errors = validate_dict_against_schema(value=config, schema=schema)
```
robons, line 253: The behaviour we want here is for the user to be *warned* about schema validation errors - however, we should still attempt to build the cube if possible.

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,331 @@
...
+        if errors:
+            return None, errors
+
+    except Exception as err:
+        print(err)
```
robons, line 258: Let's not suppress errors. Is this something you need to catch here? Can we just leave it to propagate to the top level function and be caught there? 

If we do need this try/catch here can we make the exception caught more specific and ensure that we use the `csvcubed.utils.log.log_exception` method to log the exception properly.

csvcubed/csvcubed/cli/build.py
```diff
@@ -0,0 +1,79 @@
...
+    validation_errors_file_out: Optional[Path],
+):
+
+    # Create a path to save the validation-errors.json file in the ./out directory
+
+    _logger.debug(f"CSV: {csv_path.absolute() if csv_path is not None else ''}")
+    _logger.debug(f"qube-config.json: {config.absolute() if config is not None else ''}")
+
+    if config:
+        cube, json_schema_validation_errors = get_cube_from_config_json(config, csv_path)
```
robons, line 37: I think there's some misunderstanding here. We don't want an either/or approach to convention vs configuration - we want users to be able to mix and match convention and configuration. You should be able to rely on some columns using the conventional approach and other columns in the same cube to be configured in the config.json.

csvcubed/csvcubed/readers/configdeserialiser.py
robons: I thought you'd gotten rid of this?

csvcubed/csvcubed/models/cube/qb/components/attribute.py
```diff
@@ -139,7 +139,7 @@ def from_data(
         description: Optional[str] = None,
         parent_attribute_uri: Optional[str] = None,
         source_uri: Optional[str] = None,
-        is_required: bool = False,
+        is_required: Optional[bool] = False,
```
robons, line 142: Not sure why you've made this change?

csvcubed/csvcubed/models/cube/qb/catalog.py
```diff
@@ -22,18 +22,24 @@ class CatalogMetadata(CatalogMetadataBase, UriIdentifiable):
     creator_uri: Optional[str] = field(default=None, repr=False)
     publisher_uri: Optional[str] = field(default=None, repr=False)
     landing_page_uris: list[str] = field(default_factory=list, repr=False) 
-    theme_uris: list[str] = field(default_factory=list, repr=False)
-    keywords: list[str] = field(default_factory=list, repr=False)
+    theme_uris: Optional[list[str]] = field(default_factory=list, repr=False)
```
robons, line 25: Do we need to mark these as `Optional`, won't they automatically be created as an empty list if no value has been provided by the user?

csvcubed/csvcubed/utils/json.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    Loads a resource from a URI using the requests library
+    Returns a dict of the response content or
+    Raises the Exceptions once logging them
+    """
+    try:
+        response = requests.get(uri)
+        if response.status_code != requests.codes("ok"):
+            # HTTP Get request failed - raise error
+            log.error(
```
robons, line 21: I suspect we should be raising an exception at this point.

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,421 @@
...
+    NewQbCodeListInCsvW,
+)
+from csvcubed.inputs import PandasDataTypes
+from csvcubed.utils.uri import looks_like_uri, csvw_column_name_safe
+
+T = TypeVar("T", bound=object)
+
+
+@dataclass
+class SchemaBaseClass(DataClassBase):
```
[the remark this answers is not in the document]
Giraffe-Technology: Has been moved into csvcubed/readers/v1_0
 - columnschema.py
 - mapcolumntocomponent.py

csvcubed/csvcubed/readers/mapcolumntocomponent.py
Giraffe-Technology: Updates were made following discussion so they are now separate and it uses the proposed logic to choose.

csvcubed/csvcubed/utils/validators/schema.py
Giraffe-Technology: log created at module level, no longer passed.

csvcubed/csvcubed/readers/columnschema.py
Giraffe-Technology: Done, with _get_code_list method updated.

csvcubed/csvcubed/readers/columnschema.py
Giraffe-Technology: restored

csvcubed/csvcubed/readers/columnschema.py
Giraffe-Technology: no longer present/

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+        if self.literalValuesDataType is None:
+            return ExistingQbAttribute(
+                self.uri,
+                new_attribute_values=_get_new_attribute_values(
+                    data, self.newAttributeValues
+                ),
+                is_required=self.isRequired,
+            )
+        else:
+            if self.newAttributeValues:
```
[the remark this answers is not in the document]
Giraffe-Technology: updated

csvcubed/csvcubed/readers/columnschema.py
Giraffe-Technology: fixed

csvcubed/csvcubed/readers/columnschema.py
Giraffe-Technology: fixed

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+    def map_to_qb_multi_units(
+        self, data: PandasDataTypes, column_title: str
+    ) -> QbMultiUnits:
+        return QbMultiUnits.existing_units_from_data(
+            data, csvw_column_name_safe(column_title), self.value
+        )
+
+
+@dataclass
+class UnitType(SchemaBaseClass):
```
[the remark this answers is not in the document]
Giraffe-Technology: renamed and refactored.

csvcubed/csvcubed/readers/columnschema.py
Giraffe-Technology: renamed and refactored

csvcubed/csvcubed/readers/columnschema.py
```diff
@@ -0,0 +1,447 @@
...
+
+            return QbMultiMeasureDimension(new_measures)
+
+        else:
+            raise ValueError(f"Unexpected values type: {self.values}")
+
+
+@dataclass
+class ObservationValue(SchemaBaseClass):
+    datatype: Optional[str] = None
```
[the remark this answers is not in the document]
Giraffe-Technology: done.

csvcubed/csvcubed/readers/columnschema.py
Giraffe-Technology: Updated but needs confirming

csvcubed/csvcubed/readers/columnschema.py
Giraffe-Technology: fixed

csvcubed/csvcubed/readers/configdeserialiser.py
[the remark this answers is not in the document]
Giraffe-Technology: also commented out as per your comment and confirmation with canwaf

csvcubed/csvcubed/cli/build.py
Giraffe-Technology: updated to mkdir(parents=True)

csvcubed/csvcubed/cli/build.py
Giraffe-Technology: done

csvcubed/csvcubed/cli/build.py
Giraffe-Technology: done

csvcubed/csvcubed/readers/configdeserialiser.py
Giraffe-Technology: Message updated, code refactored to '_read_and_check_csv(csv_path)' to for the csv read and checks, following changes to enable  mixed config/convention columns.

csvcubed/csvcubed/cli/build.py
Giraffe-Technology: refactored, abstracting some checks and combining the column processing to support mixed config / convention column creation.

csvcubed/csvcubed/models/cube/qb/catalog.py
Giraffe-Technology: Confirmed with canwaf - commented out.

csvcubed/csvcubed/readers/configdeserialiser.py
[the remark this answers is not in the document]
Giraffe-Technology: zapped

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,572 @@
...
+        column_type = (
+            convention_name_matches[0] if convention_name_matches else "dimension"
+        )
+        column_types.append(column_type)
+        convention_col_counts[column_type] += 1
+
+    #  Check we have the pre-requisite columns
+    if convention_col_counts["dimension"] < 1:
+        raise ValueError("The data does not contain any dimension columns")
+    if convention_col_counts["observations"] < 1:
```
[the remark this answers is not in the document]
Giraffe-Technology: zapped as didn't fit with mix n match

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,572 @@
...
+
+    # Identify column types and count
+    for i, column_name in enumerate(data.columns):
+        # Determine column type by matching column names accepted by convention or default to 'dimension'
+        convention_name_matches = [
+            standard_name
+            for standard_name, options in CONVENTION_NAMES.items()
+            if column_name.lower() in options
+        ]
+        column_type = (
```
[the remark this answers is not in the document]
Giraffe-Technology: columns new created from config in first instance, so this should be ok.

csvcubed/csvcubed/readers/configdeserialiser.py
Giraffe-Technology: removed.

csvcubed/csvcubed/cli/build.py
Giraffe-Technology: removed

csvcubed/csvcubed/models/cube/qb/components/measuresdimension.py
```diff
@@ -6,7 +6,7 @@
 """
 
 from dataclasses import dataclass
-from typing import List
+from typing import List, Union
```
robons, line 9: Can we get rtid of this unnecessary `Union` import?

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -122,7 +122,7 @@ def _metadata_from_dict(config: dict) -> "CatalogMetadata":
 
     return CatalogMetadata(
         identifier=config.get("id"),
-        title=config.get("title"),
+        title=config.get("title", ""),
```
robons, line 125: I'd say we should be raising an exception if the user hasn't provided a title. It's a required value and I really don't like it being left empty. It'll make it almost impossible to find the dataset when uploaded to PMD.

csvcubed/csvcubed/readers/v1_0/columnschema.py
```diff
@@ -0,0 +1,459 @@
...
+    csvw_column_name_safe,
+    looks_like_uri,
+)
+
+T = TypeVar("T", bound=object)
+
+
+@dataclass
+class SchemaBaseClass(DataClassBase):
+    ...
```
robons, line 54: If there's no implementation, can you ensure that it implements the `ABC` abstract base class so it isn't possible to instantiate instances of it.

csvcubed/csvcubed/readers/v1_0/columnschema.py
```diff
@@ -0,0 +1,459 @@
...
+                    code=c.code,
+                )
+                for c in new_dimension.code_list.concepts
+            ],
+        )
+
+
+@dataclass
+class ExistingDimension(SchemaBaseClass):
+    from_existing: str = ""
```
robons, line 150: I don't think `from_existing` should have a default value if it's the only property which classifies it as being an `ExistingDimension`.

csvcubed/csvcubed/readers/v1_0/columnschema.py
```diff
@@ -0,0 +1,459 @@
...
+@dataclass
+class AttributeValue(SchemaBaseClass):
+    label: str
+    description: Optional[str] = None
+    from_existing: Optional[str] = None
+    definition_uri: Optional[str] = None
+
+
+@dataclass
+class NewAttributeProperty(SchemaBaseClass):
```
robons, line 168: I don't think this class should exist - it's a hangover from the info.json v1.1 structure.

csvcubed/csvcubed/readers/v1_0/columnschema.py
```diff
@@ -0,0 +1,459 @@
...
+    comment: Optional[str] = None
+    isDefinedBy: Optional[str] = None
+    subPropertyOf: Optional[str] = None
+    newAttributeValues: Union[None, bool, List[AttributeValue]] = None
+    literalValuesDataType: Optional[str] = None
+
+
+@dataclass
+class ExistingAttribute(SchemaBaseClass):
+    from_existing: str = ""
```
robons, line 180: Again, this shouldn't have a default value since it's a required value for an `ExistingAttribute`.

csvcubed/csvcubed/readers/v1_0/columnschema.py
```diff
@@ -0,0 +1,459 @@
...
+    ) -> ExistingQbAttribute:
+        return ExistingQbAttribute(
+            self.from_existing,
+            new_attribute_values=_get_new_attribute_values(
+                data, self.values
+            ),
+            is_required=self.required,
+        )
+
+    def map_to_existing_qb_attribute(
```
robons, line 199: I don't see why you have two functions here, but this one seems like the better function to call. You should probably get rid of `map_to_qb_attribute`.

csvcubed/csvcubed/readers/v1_0/columnschema.py
```diff
@@ -0,0 +1,459 @@
...
+    definition_uri: Optional[str] = None
+    scaling_factor: Optional[float] = None
+    si_scaling_factor: Optional[float] = None
+    quantity_kind: Optional[str] = None
+
+
+@dataclass
+class ExistingUnits(SchemaBaseClass):
+    values: Union[bool, List[Unit]] = True
+    cell_uri_template: str = ""
```
robons, line 286: This shoudn't have a default value since it's required.

robons (commented):
I haven't completed this review but there are sufficient comments for you to begin working on them.

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,253 @@
...
+from csvcubed.utils.dict import get_with_func_or_none
+from csvcubed.utils.uri import uri_safe
+from csvcubed.utils.json import load_json_from_uri, read_json_from_file
+from csvcubed.utils.validators.schema import validate_dict_against_schema
+from csvcubedmodels.rdf.namespaces import GOV
+
+
+# Used to determine whether a column name matches accepted conventions
+CONVENTION_NAMES = {
+    "measures": [
```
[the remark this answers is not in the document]
Giraffe-Technology: Given the number of items there would be negligible difference but have updated to sets.

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,253 @@
...
+    """
+    Generates a Cube structure from a config.json input.
+    :return: tuple of cube and json schema errors (if any)
+    """
+    data: pd.DataFrame = _read_and_check_csv(csv_path)
+
+    # If we have a config json file then load it and validate against its reference schema
+    if config_path:
+        config = _load_resource(config_path.resolve())
+        schema = _load_resource(Path(config["$schema"]).resolve())
```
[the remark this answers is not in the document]
Giraffe-Technology: Path works, but just found .resolve() doesn't, updated _load_resource to not use urllib as schema is available is path.parts[0]

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,253 @@
...
+
+    else:
+        schema = read_json_from_file(resource_path)
+
+    return schema
+
+
+def _override_catalog_metadata_state(
+    catalog_metadata_json_file: Path, cube: QbCube
+) -> None:
```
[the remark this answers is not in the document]
Giraffe-Technology: no, removed.

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,253 @@
...
+
+def _from_config_json_dict(
+    data: pd.DataFrame, config: Optional[Dict], json_parent_dir: Path
+) -> QbCube:
+    columns: List[QbColumn] = []
+    if not config:
+        metadata: CatalogMetadata = _metadata_from_dict({})
+    else:
+        metadata: CatalogMetadata = _metadata_from_dict(config)
+        for column_title in config.get("columns", []):
```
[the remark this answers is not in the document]
Giraffe-Technology: {} creates and empty dict ?, [] creates an empty list,
No unless we are changing the Cube dataclass and 
class Cube(Generic[TMetadata], PydanticModel):
    metadata: TMetadata
    data: Optional[pd.DataFrame] = field(default=None, repr=False)
    columns: List[CsvColumn] = field(default_factory=lambda: [], repr=False)
May need to discuss this

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,253 @@
...
+        overriding_keys={k for k in catalog_metadata_dict.keys()},
+    )
+
+
+def _from_config_json_dict(
+    data: pd.DataFrame, config: Optional[Dict], json_parent_dir: Path
+) -> QbCube:
+    columns: List[QbColumn] = []
+    if not config:
+        metadata: CatalogMetadata = _metadata_from_dict({})
```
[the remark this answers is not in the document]
Giraffe-Technology: quoting documentation: "A capital case version of the csv file name"
I've updated to replace hyphens and capitalise the path stem (no extn)
raising the question of underscore handling ?

csvcubed/csvcubed/readers/configdeserialiser.py
```diff
@@ -0,0 +1,253 @@
...
+        schema_validation_errors = []
+
+    parent_path = csv_path.parent if not config_path else config_path.parent
+    cube = _from_config_json_dict(data, config, parent_path)
+
+    # Update metadata from csv where appropriate
+    cube.metadata.title = (
+        cube.metadata.title
+        if cube.metadata.title
+        else csv_path.name.upper().replace("_", " ")
```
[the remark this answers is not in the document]
Giraffe-Technology: Updated to make it clearer, although it now repeats the string formatting.
This also enabled an if statement to be removed from _from_config_json_dict.

csvcubed/csvcubed/models/cube/qb/components/measure.py
```diff
@@ -51,7 +51,7 @@ def get_default_node_serialisation_hint(self) -> RdfSerialisationHint:
 
 @dataclass
 class NewQbMeasure(QbMeasure, UriIdentifiable):
-    label: str
+    label: str = field(default="", repr=True)
```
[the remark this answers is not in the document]
Giraffe-Technology: That's not what is indicated in qbconfig-example.jsonc, 
but have updated as the documentation now indicates it's required unless from_existing is provided

csvcubed/csvcubed/readers/v1_0/columnschema.py
Giraffe-Technology: done

csvcubed/csvcubed/readers/v1_0/columnschema.py
Giraffe-Technology: removed, it was left-over from the combined New/Existing code.

csvcubed/csvcubed/readers/v1_0/columnschema.py
Giraffe-Technology: agreed - deleted

csvcubed/csvcubed/readers/configdeserialiser.py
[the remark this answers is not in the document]
robons: The `config` is a json dict of the contents of the user's `qube-config.json` file. The column mappings are a dictionary, not a list.

csvcubed/csvcubed/readers/v1_0/columnschema.py
```diff
@@ -0,0 +1,459 @@
...
+            else:
+                code_list_path = Path(self.code_list)
+                if code_list_path.is_absolute():
+                    code_list_obj = NewQbCodeListInCsvW(code_list_path)
+                else:
+                    code_list_obj = NewQbCodeListInCsvW(
+                        json_parent_dir / self.code_list
+                    )
+        elif isinstance(self.code_list, bool):
+            if self.code_list is False:
```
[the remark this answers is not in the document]
Giraffe-Technology: This is (I beleive) in line with info json, I've added a comment to help.
# The NewQbCodeList and Concepts are populated in the NewQbDimension.from_data() call
# the _get_code_list method overrides the code_list if required.

Perhaps we want to change the logic so that the NewQbCodeList is created within the _get_code_list method. - Please confirm.

csvcubed/csvcubed/readers/configdeserialiser.py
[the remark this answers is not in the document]
robons: I think both dashes and underscores should be replaced with a space.

csvcubed/csvcubed/readers/v1_0/columnschema.py
Giraffe-Technology: done.

csvcubed/csvcubed/readers/v1_0/columnschema.py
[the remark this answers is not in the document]
robons: https://github.com/GSS-Cogs/gss-utils/blob/master/gssutils/csvcubedintegration/configloaders/infojson1point1/columnschema.py#L70

So it looks like we are missing some logic (see above link) which supports the `"code_list": true` behaviour.

csvcubed/csvcubed/readers/columnschema.py
Giraffe-Technology: removed from schema, but still in QbWriter

csvcubed/csvcubed/readers/columnschema.py
Giraffe-Technology: Can you check this again please

csvcubed/csvcubed/readers/v1_0/columnschema.py
```diff
@@ -0,0 +1,459 @@
...
+    data_type: Optional[str] = None
+    required: bool = False
+    values: Union[bool, List[AttributeValue]] = True
+
+    def map_to_qb_attribute(
+        self,
+        label: str,
+        data: PandasDataTypes
+    ) -> ExistingQbAttribute:
+        return ExistingQbAttribute(
```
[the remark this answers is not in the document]
Giraffe-Technology: Can you check logic here please

csvcubed/csvcubed/writers/qbwriter.py
```diff
@@ -152,23 +152,6 @@ def _output_new_code_list_csvws(self, output_folder: Path) -> None:
 
                 code_list_writer = SkosCodeListWriter(code_list)
                 code_list_writer.write(output_folder)
-            elif isinstance(code_list, NewQbCodeListInCsvW):
```
robons, line 155: Woah, what are you deleting this stuff for?

csvcubed/csvcubed/readers/vnum.py
```diff
@@ -2,4 +2,4 @@
 
 
 class JsonSchemaVersion(Enum):
-    v1_0 = "V1_0"
\ No newline at end of file
+    v1_0 = "V1_0"
```
robons, line 5: TODO: Need to replace references to this with the stuff in `schema_versions.py`.

csvcubed/csvcubed/readers/configdeserialiser.py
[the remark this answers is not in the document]
robons: It doesn't work for URLs, it replaces `https://something` with `https:/something` which meant the code never worked to retrieve it from the web.

csvcubed/csvcubed/readers/v1_0/columnschema.py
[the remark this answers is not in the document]
robons: I've added this in.

PR #342, opened 2022-02-21
327 displaying data cube information

GDonRanasinghe:
This PR includes the following features:
1. Displaying catalog metadata.
3. Displaying DSD information and components.
4. Listing of cols with suppress output.
5. Displaying data cube code list info.
6. Unit tests.
7. Behave tests.

Typical output for `csvcubed inspect datacube.csv-metadata.json`: https://drive.google.com/file/d/1lDRqNOsNXC5sZC5YyBoktv1yi0j97cHs/view?usp=sharing

Typical output for `csvcubed inspect codelist.csv-metadata.json`: https://drive.google.com/file/d/1X3YiC_kpju1rgNO8F50-D6iDnKP1e4ye/view?usp=sharing


csvcubed-pmd/csvcubedpmd/dcatcli/pmdify.py
```diff
@@ -374,7 +374,7 @@ def _get_catalog_entry_from_dcat_dataset(csvw_graph: Graph) -> pmdcat.Dataset:
             (GROUP_CONCAT(?landingPage ; separator='|') as ?landingPages) 
             (GROUP_CONCAT(?theme; separator='|') as ?themes) 
             (GROUP_CONCAT(?keyword; separator='|') as ?keywords) 
-            ?contactPoint ?identifier
+            ?contactPoint ?identifier ?datasetContents
```
robons, line 377: This comes from another commit. Are we able to remove this commit to avoid duplication?

csvcubed/csvcubed/utils/qb/components.py
```diff
@@ -0,0 +1,41 @@
...
+
+def get_printable_component_property_type(property_type: str) -> str:
+    """
+    Produces the user-friendly name of component property type.
+
+    Member of :file:`./utils/qb/components.py`
+
+    :return: `str` - user-friendly name of component property type.
+    """
+    if ComponentPropertyType.Dimension.value in property_type:
```
robons, line 34: Can we check the whole URI matches exactly? This feels a bit loose using `in`.

csvcubed/tests/behaviour/inspectcli.feature
```diff
@@ -0,0 +1,11 @@
+# Feature: Testing inspect cli (WIP).
+
+#     Scenario: inspect cli should produce printable for a valid dataset json-ld input
+#         Given the existing test-case file "cli/inspect/dataset.csv-metadata.json"
```
robons, line 4: Once we're more comfortable with what the printed structure will looks like, it'll be good to get some behave tests working for it.

csvcubed/tests/unit/cli/inspect/test_inspectsparqlqueries.py
```diff
@@ -0,0 +1,150 @@
...
+    components = select_csvw_dsd_qube_components(
+        csvw_metadata_rdf_graph, result_dict["dataStructureDefinition"]
+    )
+
+    assert (
+        str(result_dict["dataStructureDefinition"])
+        == "file:///workspaces/csvcubed/alcohol-bulletin.csv#structure"
+    )
+    assert str(result_dict["dataSetLabel"]) == "Alcohol Bulletin"
+    assert len(components) == 17
```
robons, line 141: It would probably be useful to do a bit more in-depth testing that all of the appropriate fields come back here and that optional fields work as expected too. 

I wonder whether it might be easier to write a class to hold the information with clearly defined types instead of relying on dictionaries?

csvcubed/tests/unit/cli/inspect/test_inspectsparqlqueries.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def test_select_cols_w_supress_output():
+    csvw_metadata_json_path = _test_case_base_dir / "datacube.csv-metadata.json"
+    metadata_processor = MetadataProcessor(csvw_metadata_json_path)
+    csvw_metadata_rdf_graph = metadata_processor.load_json_ld_to_rdflib_graph()
+
+    results = select_cols_w_supress_output(csvw_metadata_rdf_graph)
+    assert len(results) == 0
```
robons, line 150: We should probably test a CSV-W where we do have at least one suppressed column.

csvcubed/csvcubed/utils/qb/components.py
GDonRanasinghe: @robons Before I was thinking that the URLs could change. However, I've found that this is not the case according to linked-data docs. So I've updated this to URI matching.

csvcubed/tests/behaviour/inspectcli.feature
GDonRanasinghe: I agree - I have this as an item in my list.

robons:
Could you add a copy of the typical output into the PR description/comment?

csvcubed/csvcubed/utils/qb/components.py
```diff
@@ -0,0 +1,79 @@
...
+
+    :return: `str` - url or relative path
+    """
+    if component_property.startswith("file://") == False:
+        return component_property
+
+    component_property = component_property.removeprefix("file://")
+    try:
+        component_property_path = Path(component_property)
+        relative_path = component_property_path.relative_to(input_file_path.parent)
```
robons, line 76: I found that [os.path.relpath](https://docs.python.org/3/library/os.path.html#os.path.relpath) does the relative paths in a more robust way than Pathlib; it does the `../` to backout to parent directory:

```python
>>> os.path.relpath("/home/something", "/chicken/burgers")
'../../home/something'
```

csvcubed/csvcubed/utils/sparql.py
```diff
@@ -4,10 +4,27 @@
...
+        return map_func(val)
 
 
 def ask(query: str, graph: Graph) -> bool:
+    """
+    Executes the given ASK query on the rdf graph.
+
+    Member of :file:`./sparql.py`.
+
+    :return: `List[ResultRow]` - List containing the results.
```
robons, line 26: This returns a bool.

csvcubed/tests/behaviour/steps/inspectcli.py
```diff
@@ -0,0 +1,117 @@
...
+from behave import *
+
+from csvcubeddevtools.behaviour.file import get_context_temp_dir_path
+from csvcubed.cli.inspect.metadatainputvalidator import MetadataValidator
+from csvcubed.cli.inspect.metadataprinter import MetadataPrinter
+from csvcubed.cli.inspect.metadataprocessor import MetadataProcessor
+
+
+def _unformat_multiline_string(string: str) -> str:
+    return "".join(s for s in string if ord(s) > 32 and ord(s) < 126)
```
robons, line 12: I can guess what this is doing - I suspect it's removing colour chars and other non-text stuff that might get printed. It'd be useful to have the exact purpose clarified though.

csvcubed/tests/unit/utils/qb/test_component.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    property_type = get_printable_component_property_type(
+        "http://purl.org/linked-data/cube#AttributeProperty"
+    )
+    assert property_type == ComponentPropertyType.Attribute.value
+
+
+def test_printable_component_property_type_measure():
+    """
+    Should return "Attribute" when the input is http://purl.org/linked-data/cube#MeasureProperty
```
robons, line 65: I agree with the behaviour but not the comment.

csvcubed/tests/unit/cli/inspect/test_metadataprocessor.py
```diff
@@ -0,0 +1,24 @@
...
+
+def test_metadata_codelist_json_ld_to_rdf_loading():
+    """
+    Metadata code list RDF graph should not be None.
+    """
+    csvw_metadata_json_path = _test_case_base_dir / "codelist.csv-metadata.json"
+    metadata_processor = MetadataProcessor(csvw_metadata_json_path)
+    csvw_metadata_rdf_graph = metadata_processor.load_json_ld_to_rdflib_graph()
+    
+    assert csvw_metadata_rdf_graph is not None
```
robons, line 24: I'd be tempted to `assert any(csvw_metadata_rdf_graph)` to ensure it's loaded some triples.

csvcubed/tests/behaviour/steps/inspectcli.py
```diff
@@ -0,0 +1,117 @@
...
+@Then("the Code List Printable should be")
+def step_impl(context):
+    assert _unformat_multiline_string(
+        context.codelist_info_printable
+    ) == _unformat_multiline_string(context.text)
+
+
+@Given('a none existing test-case file "{csvw_metadata_file}"')
+def step_impl(_,csvw_metadata_file:str):
+    pass
```
robons, line 108: Do we need this?

csvcubed/tests/behaviour/inspectcli.feature
```diff
@@ -0,0 +1,122 @@
...
+        And the Code List Printable should be
+        """
+        - The data cube has the following code lists:
+                                             Code List   Code List Label  Columns Used In
+        ./alcohol-sub-type.csv#scheme/alcohol-sub-type                   Alcohol Sub Type
+          ./alcohol-content.csv#scheme/alcohol-content                    Alcohol Content
+        ./clearance-origin.csv#scheme/clearance-origin                   Clearance Origin
+        """
+
+    Scenario: inspect cli should produce printable for code list metadata json-ld input
```
robons, line 94: I'd avoid using `cli` in the scenario descriptions since you're not actually calling the CLI here but the functions under the hood.

csvcubed/tests/behaviour/inspectcli.feature
```diff
@@ -0,0 +1,122 @@
...
+                      duties-publications-bulletin-and-factsheet).
+          
+                      *[HMRC]: HM Revenue and Customs
+                      *[UK]: United Kingdom
+        """
+        And the Data Structure Definition Printable should be
+        """
+        - The data cube has the following data structure definition:
+        - Dataset label: Alcohol Bulletin
+        - Columns with suppress output: None
```
robons, line 63: Could we move this below the `components` section? I think it's less important.

robons (commented):
Lots of good work in here! See comments.

csvcubed/csvcubed/cli/inspect/inspectsparqlmanager.py
```diff
@@ -0,0 +1,174 @@
...
+    SELECT_CODE_LISTS_AND_COLS = "select_code_lists_and_cols"
+
+
+def _get_query_string_from_file(queryType: SPARQLQueryFileName) -> str:
+    """
+    Read the sparql query string from sparql file for the given query type.
+
+    Member of :file:`./inspectsparqlmanager.py`
+
+    :return: `bool` - String containing the sparql query.
```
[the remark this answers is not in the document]
GDonRanasinghe: It is returning a `str`. I have updated.

csvcubed/csvcubed/cli/inspect/inspectsparqlmanager.py
```diff
@@ -0,0 +1,174 @@
...
+            / "inspect_sparql_queries"
+            / (queryType.value + ".sparql")
+        )
+        with open(
+            file_path,
+            "r",
+        ) as f:
+            return f.read()
+    except Exception as ex:
+        raise Exception("An error occured while reading sparql query") from ex
```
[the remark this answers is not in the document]
GDonRanasinghe: Yes a very good idea - I have updated accordingly.

csvcubed/tests/unit/utils/qb/test_component.py
GDonRanasinghe: This is now resolved.

csvcubed/tests/unit/cli/inspect/test_metadataprocessor.py
GDonRanasinghe: This is now resolved.

csvcubed/tests/behaviour/steps/inspectcli.py
GDonRanasinghe: I have made an update here - Given can be used to store the file path so that the next step dont need the path as an arg.

csvcubed/tests/behaviour/steps/inspectcli.py
GDonRanasinghe: This is now updated.

csvcubed/tests/behaviour/inspectcli.feature
GDonRanasinghe: This is now updated.

csvcubed/csvcubed/utils/qb/components.py
```diff
@@ -0,0 +1,79 @@
...
+    input_file_path: Path, component_property: str
+) -> str:
+    """
+    Produces the user-friendly property of the component property.
+
+    Member of :file:`./utils/qb/components.py`
+
+    :return: `str` - url or relative path
+    """
+    if component_property.startswith("file://") == False:
```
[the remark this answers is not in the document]
GDonRanasinghe: Yes, that is more readable too - I have updated this.

csvcubed/csvcubed/cli/inspect/inspectsparqlmanager.py
```diff
@@ -0,0 +1,174 @@
...
+    return select(
+        _get_query_string_from_file(SPARQLQueryFileName.SELECT_DSD_QUBE_COMPONENTS),
+        rdf_graph,
+        init_bindings={"dsd_uri": URIRef(dsd_uri)},
+    )
+
+
+def select_cols_w_supress_output(rdf_graph: Graph) -> List[ResultRow]:
+    """
+    Queries the columns with suppress output is true.
```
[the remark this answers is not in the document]
GDonRanasinghe: This is now updated.

csvcubed/csvcubed/cli/inspect/inspectsparqlmanager.py
```diff
@@ -0,0 +1,174 @@
...
+    """
+    results: List[ResultRow] = select(
+        _get_query_string_from_file(
+            SPARQLQueryFileName.SELECT_DSD_DATASETLABEL_AND_URI
+        ),
+        rdf_graph,
+    )
+
+    if len(results) != 1:
+        raise Exception(f"Expected 1 dcat:Dataset record, found {len(results)}")
```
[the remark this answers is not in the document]
GDonRanasinghe: I have updated this now.

csvcubed/csvcubed/cli/inspect/metadataprinter.py
```diff
@@ -50,21 +127,104 @@ def gen_dsd_info_printable(self) -> str:
...
+            map(
+                lambda component: {
+                    "componentProperty": get_printable_component_property(
+                        self.csvw_metadata_json_path,
+                        component["componentProperty"],
+                    ),
+                    "componentPropertyLabel": none_or_map(
+                        component.get("componentPropertyLabel"), str
+                    )
+                    or "",
```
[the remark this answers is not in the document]
GDonRanasinghe: @robons I am using the black formatter in vs code. And it is doing this line breaking. Would you still like anything changed here?

csvcubed/csvcubed/cli/inspect/inspect.py
```diff
@@ -67,13 +73,19 @@ def _generate_printables(
...
-        metadata_printer.gen_type_info_printable(csvw_type),
-        metadata_printer.gen_metadata_info_printable(),
-        metadata_printer.gen_dsd_info_printable(),
-        metadata_printer.gen_codelist_info_printable(),
-        metadata_printer.gen_headtail_printable(),
-        metadata_printer.gen_valcount_printable(),
+        metadata_printer.gen_type_info_printable(),
+        metadata_printer.gen_catalog_metadata_printable(),
+        metadata_printer.gen_dsd_info_printable()
+        if csvw_type == CSVWType.QbDataSet
```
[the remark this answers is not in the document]
GDonRanasinghe: Improved readability here.

csvcubed/csvcubed/models/inspectsparqlresults.py
```diff
@@ -0,0 +1,209 @@
...
+)
+
+
+@dataclass()
+class CatalogMetadataSparqlResult:
+    """
+    Model to represent result of the `select_csvw_catalog_metadata` sparql query.
+    """
+
+    sparql_result: ResultRow
```
robons, line 29: I was more thinking that we'd do the mapping outside of this model so that this model could be used for more than just printing strings in the future.

csvcubed/csvcubed/models/inspectsparqlresults.py
```diff
@@ -0,0 +1,209 @@
...
+        formatted_landing_pages: str = get_printable_list_str(self.landing_pages)
+        formatted_themes: str = get_printable_list_str(self.themes)
+        formatted_keywords: str = get_printable_list_str(self.keywords)
+        formatted_description: str = self.description.replace(linesep, f"{linesep}\t\t")
+        return f"{linesep}\t- Title: {self.title}{linesep}\t- Label: {self.label}{linesep}\t- Issued: {self.issued}{linesep}\t- Modified: {self.modified}{linesep}\t- License: {self.license}{linesep}\t- Creator: {self.creator}{linesep}\t- Publisher: {self.publisher}{linesep}\t- Landing Pages: {formatted_landing_pages}{linesep}\t- Themes: {formatted_themes}{linesep}\t- Keywords: {formatted_keywords}{linesep}\t- Contact Point: {self.contact_point}{linesep}\t- Identifier: {self.identifier}{linesep}\t- Comment: {self.comment}{linesep}\t- Description: {formatted_description}"
+
+    def __post_init__(self):
+        result_dict = self.sparql_result.asdict()
+
+        self.title: str = str(result_dict["title"])
```
robons, line 42: I'm not sure that these get counted as dataclass fields since they're not defined in the standard location.

csvcubed/csvcubed/models/inspectsparqlresults.py
robons: Perhaps we could bring the attribute definitions outside of `__post_init__` and then have a static method which takes a `ResultRow` and creates the `CatalogMetadataSparqlResult`? That way we're not exposing the `ResultRow` to anywhere that doesn't need to know about it - people can just reference the individual properties then.

robons (approved):
Looks good to me. 

The only suggestion I would make is not to use `Model` in the name of models. Perhaps `Result` would make it clear it's the result of a query? And when it comes to the Catalog Metadata result, it's possible you could reuse the existing `CatalogMetadata` model.

GDonRanasinghe:
> 

@robons Sure, I will make sure to do this change in the next PR.

PR #343, opened 2022-02-21
Canwaf/pandas-upgrade

canwaf:
Release the pandas versions to anything greater than 1.3 but less than 2.0. Since we're not doing anything fancy and their roadmap doesn't involve breaking the very basic functions we use we should be fine until we're not.


robons (approved):
I approve.

robons:
Don't forget you'll need to go via the whole releast proceedure to get this updated in gss-utils.

PR #351, opened 2022-03-02
func to handle preset templates

MuazzamChaud:


csvcubed/tests/unit/readers/test_preset_column_templates.py
```diff
@@ -0,0 +1,32 @@
...
+    apply_preconfigured_values_from_template(column_config, "V1_0")
+    assert column_config.get("column_uri_template") == "http://....../{+<column_name>}"
+
+def test_func_accepts_dict_without_template():
+    column_config = {
+        "label" : "year",
+    }
+    assert "label" in column_config
+    assert "column_uri_template" not in column_config
+    assert (apply_preconfigured_values_from_template(column_config, "v1_0")) == None
```
robons, line 19: This one should raise an exception IMO since it can't find the location of the lookup JSON file (lower-case `v`).

robons: I suppose it doesn't need to fetch the file if it doesn't have a template.

csvcubed/csvcubed/readers/V1_0/preconfiguredtemplates.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import Dict, Any
+
+def apply_preconfigured_values_from_template(column_config: Dict[str, Any], version_module_path: str) -> None:
+    # if column_config doesn't have the `from_template` property, just terminate the function now
+    if "from_template" not in column_config:
+        return
+        
+    # if column_config has `from_template` property, extract that value
+    # remove the `from_template` property
+    else:
```
robons, line 19: You don't need an `else` statement here since the other condition exits from the function.

csvcubed/csvcubed/readers/V1_0/preconfiguredtemplates.py
```diff
@@ -0,0 +1,46 @@
...
+        template_value = column_config["from_template"]
+        del column_config["from_template"]
+        logging.debug(column_config)
+
+    # given the `from_template` value, look up the template in the git repo
+    url =  f"https://raw.githubusercontent.com/GSS-Cogs/csvcubed/main/csvcubed/csvcubed/readers/{version_module_path}/templates/preset_column_config.json"
+    logging.debug(url)
```
robons, line 26: Can you add some text to the debug message to make the context clear? It's probably a good idea in all of the debug statements you're using.

csvcubed/tests/unit/readers/test_preset_column_templates.py
```diff
@@ -0,0 +1,32 @@
+import pytest
+from csvcubed.readers.V1_0.preconfiguredtemplates import apply_preconfigured_values_from_template
+
+
+def test_func_accepts_dict_with_template():
+    column_config = {
+        "label" : "year",
+        "from_template" : "year"
+    }
+    apply_preconfigured_values_from_template(column_config, "V1_0")
```
robons, line 10: Can we move the `V1_0` stuff into an enum so it's hard to mess it up?

csvcubed/csvcubed/models/cube/qb/components/attribute.py
```diff
@@ -72,20 +71,6 @@ class ExistingQbAttribute(QbAttribute):
     is_required: bool = field(default=False, repr=False)
     arbitrary_rdf: List[TripleFragmentBase] = field(default_factory=list, repr=False)
 
-    @validator("new_attribute_values")
-    def _validate_concepts_non_conflicting(
```
robons, line 76: Something gone wrong here, you're removing code that needs to be in there.

csvcubed/csvcubed/utils/cache.py
```diff
@@ -1,6 +1,7 @@
 from requests_cache import CachedSession
 
-def session(url):
+def start_session(url):
```
robons, line 3: I think it'd be better if you just exposed the `CachedSession` instead of making this function available. Code elsewhere can then call `cached_session.get("http://my-url")` quite naturally.

csvcubed/csvcubed/readers/preconfiguredtemplates.py
```diff
@@ -0,0 +1,78 @@
...
+
+def get_propeties_from_template_file(template_file: str, version_module_path: str) -> dict:
+    """
+    Given the file path to the template, read in all the propeties of that particular template
+    """
+    template_template_lookup_url = f"https://raw.githubusercontent.com/GSS-Cogs/csvcubed/main/csvcubed/csvcubed/readers/{version_module_path}/templates/{template_file}"
+    template_response = session.get(template_template_lookup_url)
+    logging.critical(template_response)
+    if template_response.status_code != 200:
+        logging.warning("Status code: %s.", template_response.status_code)
```
robons, line 44: Maybe we could print the whole response in a debug statement at the same time to help figure out what went wrong?

csvcubed/csvcubed/readers/preconfiguredtemplates.py
```diff
@@ -0,0 +1,78 @@
...
+    template_file = template_lookup.get(template_value)
+
+    return template_file
+
+
+def get_propeties_from_template_file(template_file: str, version_module_path: str) -> dict:
+    """
+    Given the file path to the template, read in all the propeties of that particular template
+    """
+    template_template_lookup_url = f"https://raw.githubusercontent.com/GSS-Cogs/csvcubed/main/csvcubed/csvcubed/readers/{version_module_path}/templates/{template_file}"
```
robons, line 40: `template_template`? Been using refactoring tools to do renaming?

csvcubed/csvcubed/readers/preconfiguredtemplates.py
```diff
@@ -0,0 +1,78 @@
...
+    template_response = session.get(template_template_lookup_url)
+    logging.critical(template_response)
```
robons, line 42: We should remove this `critical` log here - I presume you were just using it for debugging.

csvcubed/csvcubed/readers/preconfiguredtemplates.py
```diff
@@ -0,0 +1,78 @@
...
+        logging.warning("Status code: %s.", template_lookup_response.status_code)
+
+    try:
+        template_lookup = template_lookup_response.json()
+    except JSONDecodeError as e:
+        raise Exception(f"Could not access template lookup file at {template_lookup_url}") from e
+
+    template_file = template_lookup.get(template_value)
+
+    return template_file
```
robons, line 33: Technicaly the return type here should be `Optional[str]` since `dict.get(...)` will return `None` if it can't find the key.

Perhaps raise an exception if the `template_file` (path) is `None` and let the user know that we couldn't find the template they were looking for?

csvcubed/csvcubed/readers/preconfiguredtemplates.py
```diff
@@ -0,0 +1,78 @@
...
+from csvcubed.utils.cache import session
+
+def get_template_file_from_template_lookup(template_value: str, version_module_path: str) -> str:
+    """
+    Given the `from_template` value, look up the template in the git repo
+    """
+
+    template_lookup_url =  f"https://raw.githubusercontent.com/GSS-Cogs/csvcubed/main/csvcubed/csvcubed/readers/{version_module_path}/templates/preset_column_config.json"
+    template_lookup_response = session.get(template_lookup_url)
+    logging.debug("The template is at: %s", template_lookup_url)
```
robons, line 21: I wouldn't say this is the template file, more the template lookup/index file.

csvcubed/csvcubed/readers/preconfiguredtemplates.py
```diff
@@ -0,0 +1,78 @@
...
+   
+    if template_lookup_response.status_code != 200:
+        logging.warning("Status code: %s.", template_lookup_response.status_code)
```
robons, line 24: let's log debug the response here to help us debug if anything goes wrong.

csvcubed/csvcubed/readers/preconfiguredtemplates.py
```diff
@@ -0,0 +1,78 @@
...
+        template_lookup = template_lookup_response.json()
+    except JSONDecodeError as e:
+        raise Exception(f"Could not access template lookup file at {template_lookup_url}") from e
+
+    template_file = template_lookup.get(template_value)
+
+    return template_file
+
+
+def get_propeties_from_template_file(template_file: str, version_module_path: str) -> dict:
```
robons, line 36: This function should probably be private and so start with a `_` in the name. This is true of `get_template_file_from_template_lookup` too.

csvcubed/csvcubed/readers/vnum.py
```diff
@@ -0,0 +1,12 @@
+from enum import Enum
+
+
+class JsonSchemaVersion(Enum):
+    v1_0 = "V1_0"
+
+def do_something(version: JsonSchemaVersion) -> None:
```
robons, line 7: Can we remove this?

csvcubed/tests/unit/readers/test_preset_column_templates.py
```diff
@@ -0,0 +1,56 @@
...
+    apply_preconfigured_values_from_template(column_config, JsonSchemaVersion.v1_0.value)
+    assert column_config.get("column_uri_template") == "http://....../{+<column_name>}"
+
+def test_func_accepts_dict_without_template():
+    column_config = {
+        "label" : "year",
+    }
+    assert "label" in column_config
+    assert "column_uri_template" not in column_config
+    assert (apply_preconfigured_values_from_template(column_config, JsonSchemaVersion.v1_0.value)) == None
```
robons, line 20: I don't think we need to assert anything here.

Just do something like this:

```python
from copy import deepcopy
column_config_copy = deepcopy(column_config)
...
apply_preconfigured_values_from_template(...)
assert column_config == column_config_copy
```

This will ensure that no changes have been made to the `column_config`.

robons (commented):
Good work. I've left a few suggested improvements.

csvcubed/csvcubed/readers/preconfiguredtemplates.py
```diff
@@ -0,0 +1,81 @@
...
+        raise Exception(f"Could not access template lookup file at {template_lookup_url}") from e
+
+    template_file = template_lookup.get(template_value)
+    if template_file == None:
+        raise Exception("Couldn't find template your looking for.")
+
+    return template_file
+
+
+def _get_propeties_from_template_file(template_file: str, version_module_path: str) -> dict:
```
mikeAdamss, line 39: can we apply the comments on `_get_template_file_from_template_lookup` to this function as well please.

csvcubed/csvcubed/readers/preconfiguredtemplates.py
```diff
@@ -0,0 +1,81 @@
...
+    del column_config["from_template"]
+
+    # given the `from_template` value, look up the template in the git repo
+    template_file = _get_template_file_from_template_lookup(template_value, version_module_path)
+
+    # given the file path to the template, read in all the propeties of that particular template
+    fetch_template = _get_propeties_from_template_file(template_file, version_module_path)
+
+    # insert values from column_config (as long as the user hasn't provided an overriding value for them)
+    for propety in fetch_template:
```
mikeAdamss, line 78: `property` is a reserved word in python, we probably shouldn't just misspell it to avoid the clash 😄 , I'd go with something like `template_property` maybe.

mikeAdamss:
@MuazzamChaud - I missed a forward slash, that `TEMPLATE_BASE_URL` should end with `templates/{}` not `templates{}`

csvcubed/csvcubed/readers/preconfiguredtemplates.py
```diff
@@ -0,0 +1,81 @@
...
+        raise HTTPError(f'Unable to get from url {template_lookup_url}. Status code: {template_lookup_response.status_code}')
+
+    try:
+        template_lookup = template_lookup_response.json()
+    except JSONDecodeError as e:
+        raise Exception(f"Could not decode response \n{template_lookup_response}\n from {template_lookup_url}") from e
+
+    template_file = template_lookup.get(template_value)
+    if not template_file:
+        raise Exception("Couldn't find template your looking for.")
```
robons, line 34: Could we add the name of the template the user provided into this error message? It'll make it easier for users to figure out what they did wrong.

csvcubed/tests/unit/readers/test_preset_column_templates.py
```diff
@@ -0,0 +1,87 @@
...
+
+TEMPLATE_BASE_URL = "https://raw.githubusercontent.com/GSS-Cogs/csvcubed/main/csvcubed/csvcubed/readers/{}/templates/{}"
+
+def test_func_accepts_dict_with_template():
+    column_config = {
+        "label" : "year",
+        "from_template" : "year"
+    }
+    apply_preconfigured_values_from_template(column_config, JsonSchemaVersion.v1_0.value)
+    assert column_config.get("column_uri_template") == "http://....../{+<column_name>}"
```
robons, line 21: Can you create an issue to fill in some real-world values here, please?

PR #364, opened 2022-03-08
Updates to project root path finding

GDonRanasinghe:
The `_get_query_string_from_file` now finds the project root by using a variable defined at the `definitions.py` file. This file is at the project root.

### DIFF
--- a/csvcubed/definitions.py
+++ b/csvcubed/definitions.py
@@ -1,0 +1,7 @@
+import os
+
+"""
+Defines a variable to the Project Root.
+Ref: https://stackoverflow.com/questions/25389095/python-get-path-of-root-project-structure/25389715#25389715
+"""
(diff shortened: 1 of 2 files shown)


csvcubed/definitions.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+"""
+Defines a variable to the Project Root.
+Ref: https://stackoverflow.com/questions/25389095/python-get-path-of-root-project-structure/25389715#25389715
+"""
+ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
```
robons, line 7: Can you convert this to a `pathlib.Path` type object?

robons (approved):
Just a small change to going the `pathlib.Path` conversion inside the `definitions` module if you could.

PR #370, opened 2022-03-23
Loading `tableSchema` file dependencies

robons:
This is necessary to ensure we can properly query the metadata surrounding code lists as part of the `csvcubed inspect` command.


csvcubed/csvcubed/cli/inspect/inspect_sparql_queries/select_single_unit_from_dsd.sparql
```diff
@@ -3,14 +3,14 @@ PREFIX csvw: <http://www.w3.org/ns/csvw#>
...
-    ?column csvw:propertyUrl  "http://purl.org/linked-data/sdmx/2009/attribute#unitMeasure"^^csvw:uriTemplate;
+    ?column csvw:propertyUrl "http://purl.org/linked-data/sdmx/2009/attribute#unitMeasure"^^csvw:uriTemplate;
             csvw:valueUrl ?unitUriTemplate.
 
     BIND(str(?unitUriTemplate) as ?unitUri).
 
 
     OPTIONAL {
-        ?unit rdfs:titles ?unitLabel.
+        ?unit rdfs:label ?unitLabel.
```
robons, line 13: @GDonRanasinghe, do we have a unit test for this sparql query? If so we should be ensuring that we can match on the unitLabel where appropriate. I only noticed this after I started playing with some inputs.

GDonRanasinghe (approved):
The PR looks good.

csvcubed/csvcubed/cli/inspect/inspect_sparql_queries/select_single_unit_from_dsd.sparql
GDonRanasinghe: @robons I had a test for this but was commented out. I have enabled this test now and it assets for the uri and label.

PR #371, opened 2022-03-23
Mike/issue 369

mikeAdamss:
fixes a bug where we were missing `url` and `tableSchema` fields where they were embedded within a `tables` field., i.e ones like this: https://ci.ukstats.dev/job/GSS_data/job/Trade/job/csvcubed/job/HMRC-alcohol-bulletin/job/HMRC-alcohol-bulletin/lastSuccessfulBuild/artifact/outputs/alcohol-sub-type.csv-metadata.json 

Also added pytest library [pytest-recording](https://pypi.org/project/pytest-recording/) so we can safely test against frozen response fixtures.

the included yml files are http responses as files that are "played" into the tests.

### DIFF
--- a/csvcubed-pmd/csvcubedpmd/csvwcli/pull.py
+++ b/csvcubed-pmd/csvcubedpmd/csvwcli/pull.py
@@ -69,7 +69,14 @@ def _get_csvw_dependencies_relative_to_metadata_file(
     metadata_file: URLOrPath,
 ) -> Set[str]:
     def _get_raw_dependencies(table_group: dict) -> Iterable[str]:
+
+        # Embedded tables
         tables = table_group.get("tables", [])
+
+        # If none, assume csvw represents a single table
+        if not any(tables):
+            tables = [table_group]
+
         for table in tables:
             table_url = table.get("url")
             schema = table.get("tableSchema")
--- a/csvcubed-pmd/tests/unit/csvwcli/test_pull.py
+++ b/csvcubed-pmd/tests/unit/csvwcli/test_pull.py
@@ -10,6 +10,7 @@ from csvcubedpmd.csvwcli.pull import (
 _test_cases_dir = get_test_cases_dir()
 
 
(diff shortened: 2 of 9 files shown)


csvcubed-pmd/tests/unit/csvwcli/test_pull.py
```diff
@@ -74,5 +79,14 @@ def test_get_relative_dependencies_excludes_absolute_dependencies():
     assert len(dependencies) == 0
 
 
+@pytest.mark.vcr
+def test_get_tableschema_dependencies():
+    dependencies = _get_csvw_dependencies_relative(
+        "https://ci.ukstats.dev/job/GSS_data/job/Trade/job/csvcubed/job/HMRC-alcohol-bulletin/job/HMRC-alcohol-bulletin/lastSuccessfulBuild/artifact/outputs/alcohol-sub-type.csv-metadata.json"
```
robons, line 85: This URL will cease to work at some point in the future. I presume that's why you've added VCR on here?

csvcubed-pmd/csvcubedpmd/csvwcli/pull.py
```diff
@@ -69,7 +69,14 @@ def _get_csvw_dependencies_relative_to_metadata_file(
     metadata_file: URLOrPath,
 ) -> Set[str]:
     def _get_raw_dependencies(table_group: dict) -> Iterable[str]:
+
+        # Embedded tables
         tables = table_group.get("tables", [])
+
+        # If none, assume csvw represents a single table
```
robons, line 76: Looks good!

csvcubed-pmd/tests/unit/csvwcli/test_pull.py
mikeAdamss: yeah exactly, though i prefer fixtures for all urls (where sensible), quite happy with that pyest plugin tbh, nice and simple.

PR #372, opened 2022-03-23
Charles rendle/issue229 Bug Reporting - User Guidance

CharlesRendle:
Draft PR as current issue form does not display as intended. At the moment, I am assuming it is a fault in my YAML syntax. 

Currently, PR will add

- Bug reporting documentation page with guidelines and error logging details
- A template for issue reporting forms

I still need to verify that when an issue is submitted, a simple github issue is created. I expect a resulting issue to look something [like this](https://github.com/pandas-dev/pandas/issues/46392) but with the template from `bug-report.yml`.


CharlesRendle:
Removed the github issue forms in favour of standard markdown templates. When adding templates, github does not allow you to commit the new markdown files to an existing branch only to open a PR in a new branch. There is a [supplementary PR](https://github.com/GSS-Cogs/csvcubed/pull/393) which will add these files and ideally should be approved simultaneously with this one.

Have also added a template and section for requesting further guidance from the team.

canwaf (approved):
Another good PR, thanks for doing the documentation. Thanks @CharlesRendle.

Issue #337, opened 2022-02-16
Change handling of NA as a string


canwaf:
pandas' `read_csv()` method changes `na` (case-insensitive) to a null value by default. The list of values which are interpreted as null (or `np.nan`) are at https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html. This behaviour needs to be supressed across the entire code base as users should be able to provide any key they would like while generating codelists, etc.

In CSV-W the specification states that the only way to convey a true null in the csv is through a zero-length string (i.e. `,,`).

- [x] Ensure this behaviour globally using system-configuration (first perference), or create a utility function to perform this behaviour within csvcubed and pull out all direct calls to pandas (second choice)
- [ ] Update the documentation that explains how the input csvs for both `infojson2csvqb` and `csvcubed` must be formatted with nulls as zero-length strings. For the csvcubed one make sure it ends up in the same documentation page that we'll be creating in Issue #274.
- [x] Make sure you do some unit/behaviour tests on this.

santhosh-thangavel:
Dataset:

https://github.com/GSS-Cogs/family-trade/blob/master/datasets/ONS-UK-SA-Trade-in-goods/main.py

Codelist:

https://github.com/GSS-Cogs/ref_common/blob/pmd4/reference/codelists/cl-area.csv (Label Namibia with Notation “NA” in the codelists)

Issue:

Dimension value in question is “Namibia” which has the country code “NA” under “ONS Partner Geography” dimension.
In the codelists, Notation for label Namibia is “NA”

The Jenkins build fails as the notation “NA” is considered as missing values in the dimension. Looks like “NA” is considered as np.nan which shouldn’t be.

Temporary fix:

Added new Label “Namibia” which has Label “NAM” in the codelist.

mikeAdamss:
Go with the second option Andrew is presenting here please. Having a python library changing system preferences  to alter the behaviour of another python library is a bit sketchy.

When you do this be sure to write a function that can propagate kwargs please, something (vaguely..... along the lines of .... don't take this as gospel):

```python
def get_pandas_dataframe(csv: Path, **kwargs) -> pd.DataFrame:

	# tweak **kwargs somehow to get the NA behaviour we want
	
	 return pandas.read_csv(csv, **kwargs)
	
``` 

Basically, don't lock the signature down, we'll probably want the ability to nuance the behaviour of `get_pandas_dataframe` in places.

[...]

PR #374, opened 2022-03-24
Muazzam chaud/issue337

MuazzamChaud:
#337


csvcubed/tests/unit/utils/test_pandas_na_filter.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+
+from csvcubed.utils.pandas_na_filter import refiltered_pandas_read_csv
+
+csv_path = "/workspaces/csvcubed/csvcubed/tests/test-cases/utils/pandas_na_filter/code-list.csv"
+
+def test_na_filter_works():
+    pd_table = refiltered_pandas_read_csv(csv_path)
```
robons, line 8: Could you make sure to add some comments describing what the test is ensuring?

csvcubed/tests/unit/utils/test_pandas_na_filter.py
```diff
@@ -0,0 +1,31 @@
...
+    assert pd_table['Description'].isnull().values.any()
+    assert pd_table['Description'].isnull().sum() == 13
+
+def test_NaN_value_in_Parent_Notation_column_is_not_typr_string():
+    pd_table = refiltered_pandas_read_csv(csv_path)
+    assert pd_table['Parent Notation'].isnull().values.any()
+    assert pd_table['Parent Notation'].isnull().sum() == 1
+
+if __name__ == "__main__":
+    pytest.main()
```
robons, line 31: This is against the file above it because I can't comment on the real issue, but the src/gssutils module shouldn't exist in this repo.

csvcubed/tests/unit/utils/test_pandas_na_filter.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+
+from csvcubed.utils.pandas_na_filter import refiltered_pandas_read_csv
+
+csv_path = "/workspaces/csvcubed/csvcubed/tests/test-cases/utils/pandas_na_filter/code-list.csv"
```
robons, line 5: This URL won't work for developers who aren't using docker. Don't forget about the function you can use which finds the `test-cases` directory for you.

csvcubed/csvcubed/utils/pandas_na_filter.py
```diff
@@ -0,0 +1,17 @@
+import pandas as pd
+
+from pathlib import Path
+
+def refiltered_pandas_read_csv(csv: Path) ->  pd.DataFrame:
```
robons, line 5: Can we just call this `read_csv`?

csvcubed/csvcubed/utils/pandas_na_filter.py
```diff
@@ -0,0 +1,17 @@
...
+
+    specified_na_values = _get_new_na_values()
```
robons, line 7: This doesn't need to be a function, it could just be referring to a private constant defined in this module.

csvcubed/csvcubed/utils/pandas_na_filter.py
```diff
@@ -0,0 +1,17 @@
...
+    df = pd.read_csv(csv, keep_default_na = False, na_values = specified_na_values)
+
+    return df
+
+def _get_new_na_values():
+    specified_na_values = {
+    "",
+    "NaN"
```
robons, line 15: I thought we didn't want `NaN` to be treated as a NaN value? Let's just leave it as blank values being treated as NaN.

csvcubed/csvcubed/utils/pandas_na_filter.py
robons: And maybe just put it inside the existing `pandas` utils module that I believe we have (or had).

csvcubed/csvcubed/utils/pandas_na_filter.py
```diff
@@ -0,0 +1,17 @@
+import pandas as pd
```
robons, line 1: Don't forget we require a title for the module that sphinx can pick up.

https://github.com/GSS-Cogs/csvcubed/blob/main/docs/developer.md#api-documentation

csvcubed/tests/unit/utils/test_pandas.py
```diff
@@ -0,0 +1,51 @@
...
+def test_na_filter_works():
+    """
+    running the 'read_csv' command and checking that some of the default na-values
+    don't get transformed to NaN values.
+    """
+    pd_table = read_csv(csv_path)
+    assert "nan" in pd_table['Description'].values
+    assert "null" in pd_table['Description'].values
+    assert "#N/A" not in pd_table['Description'].values
+    assert "null"  not in pd_table['Parent Notation'].values
```
robons, line 21: So this is saying that `null` is treated as a NaN by pandas?

robons (approved):

Issue #378, opened 2022-03-28
Support the removal of `.csv` from CSV-W generated URIs


robons:
From @rossbowen:

> can we get the .csv dropped from URIs in csvcubed, or maybe exposed as an option? We’re probably going to start work on refreshing the climate data soon (which does give an opportunity to start using csvcubed) but I’d have a preference for dataset URIs to be similar (i.e. not with the .csv ) to those coined last year

The reason we include the `.csv` on the end of URLs generated inside the CSV-W is so that when a user uploads a CSV-W to their wordpress (or similar) site, the URIs should dereference somewhere sensible (i.e. to a file which exists on their server and is related to the data.)

As part of this task, could you please:

- [x] Add a *URL style* property to the `Cube` class' constructor in csvcubed which allows a user to specify the style of URLs that csvcubed should generate.
  - The type of this property should be an enum which currently has values similar to `Standard` and `WithoutFileExtensions`
- [x] (gss-utils) Add the ability to specify a value in the info.json v1.1 format which allows the user to configure the url style generated by csvcubed.  
   - This should be picked up by the `infojson2csvw build` command and used to configure the `Cube` using the new parameter you've added above. 
   - N.B. you must update the JSON schema as well as the code.
- [x] Write unit/behaviour tests for everything
- [x] Update the appropriate documentation in gss-utils to make reference to this new functionality.

goofballLogic:
https://github.com/GSS-Cogs/gss-utils/issues/356

goofballLogic:
Part of this is: https://github.com/GSS-Cogs/family-schemas/issues/7

PR #383, opened 2022-03-31
Allow configuration of URI style for cubes

goofballLogic:
Part of #378


csvcubed/csvcubed/writers/skoscodelistwriter.py
```diff
@@ -145,11 +147,12 @@ def _get_csvw_table_schema(self) -> dict:
     def _get_csvw_metadata(self) -> dict:
         scheme_uri = self._new_uri_helper.get_scheme_uri()
         additional_metadata = self._get_catalog_metadata(scheme_uri)
+        url = self._new_uri_helper._get_identifier_for_document()
 
         return {
             "@context": "http://www.w3.org/ns/csvw",
             "@id": scheme_uri,
-            "url": self.csv_file_name,
+            "url": url,
```
robons, line 155: This needs to point at the true path of the CSV file, not the URI that we are using as an identifier for it.

Also on a side-note: `_get_identifier_for_document` is (by convention) a private function since it starts with a `_`.

csvcubed/tests/behaviour/steps/qbwriter.py
```diff
@@ -735,3 +743,89 @@ def step_impl(context, cube_name: str):
...
+            o
+            for (p, o) in g.predicate_objects()
+            if isinstance(o, URIRef)
+            and not (str(p) in predicate_whitelist)
+            and not (str(o).startswith(tuple(object_prefix_whitelist)))
+        ]
+    )
+    assertURIStyles(uri_style, uriRefObjects)
+
+    uriLiteralObjects = set(
```
robons, line 808: You can do set comprehensions just like list comprehensions in python, i.e. switch the `[ ... ]` for `{ ... }` and you don't need to convert to a set.

csvcubed/tests/behaviour/steps/qbwriter.py
```diff
@@ -735,3 +743,89 @@ def step_impl(context, cube_name: str):
...
+        cube_name, None, URIStyle[uri_style]
+    )
+
+
+def assertURIStyle(uri_style: URIStyle, temp_dir: Path, csv_file_name: str):
+    metadataFilePath = temp_dir.joinpath(f"{csv_file_name}-metadata.json")
+    g = Graph()
+    g.parse(metadataFilePath)
+
+    predicate_whitelist = [
```
robons, line 774: I see what you've done here but I do worry these whitelists will make the tests fairly brittle if and when we add new RDF into the CSV-Ws in future. 

I think the way I would have approached this problem is to look for *relative URIs* which are defined as part of the graph - these are the new resources we're creating which use the newly generated URIs.

RDFlib isn't happy with relative URIs so it converts them to absolute URIs using the `metadataFilePath` in this instance. Now we've had some problems with the way that rdflib converts file paths into URIs (where Jenkins puts things like `@2` in folder names/paths), so I'd recommend setting `publicID="http://prefix-for-relative-uris"` in your call to `g.parse` which would mean that all of your relative URIs would end up with `http://prefix-for-relative-uris` prepended. Then you can filter by URIs that begin with that and ensure that they all either have or don't have the `.csv` extension on the end.

-------

Btw, `metadataFilePath` betrays your allegiance for other languages ;) would be `metadata_file_path` by convention in python.

csvcubed/tests/behaviour/cube.feature
```diff
@@ -28,9 +28,9 @@ Feature: Cube!
...
+        {"csv_column_title": "Dim-0", "structural_definition": {"label": "Non Dimension", "description": None, "code_list": {"metadata": {"title": "Non Dimension", "identifier": None, "summary": None, "description": None, "creator_uri": None, "publisher_uri": None, "landing_page_uris": [], "theme_uris": [], "keywords": [], "dataset_issued": None, "dataset_modified": None, "license_uri": None, "public_contact_point_uri": None, "uri_safe_identifier_override": None, }, "concepts": [{"label": "1", "code": "1", "parent_code": None, "sort_order": None, "description": None, "uri_safe_identifier_override": None}, {"label": "2", "code": "2", "parent_code": None, "sort_order": None, "description": None, "uri_safe_identifier_override": None}], "arbitrary_rdf": [], "uri_style": None}, "parent_dimension_uri": None, "source_uri": None, "range_uri": None, "uri_safe_identifier_override": None, "arbitrary_rdf": []}, "csv_column_uri_template": None, "uri_safe_identifier_override": None},
```
robons, line 31: Yeah, this test isn't my favourite approach, sorry about the brittle nature.

robons (commented):
I'm impressed you managed to get this far without more support.

I've left some comments. Let me know if you need help understanding anything.

csvcubed/tests/behaviour/steps/qbwriter.py
goofballLogic: ooo nice

csvcubed/tests/behaviour/steps/qbwriter.py
goofballLogic: ah damn i missed one - was a few days in when i noticed the snake casing

goofballLogic: ok will switch to that approach. I noticed it was the relative ones I wanted but it felt a bit dirty looking for what rdflib happend to convert them too. Using publicID makes it a bit better.

csvcubed/csvcubed/models/cube/qb/components/dimension.py
```diff
@@ -96,14 +97,15 @@ def from_data(
         range_uri: Optional[str] = None,
         uri_safe_identifier_override: Optional[str] = None,
         arbitrary_rdf: List[TripleFragmentBase] = [],
+        uri_style: Optional[URIStyle] = None,
```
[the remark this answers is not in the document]
goofballLogic: done

csvcubed/csvcubed/writers/urihelpers/skoscodelist.py
```diff
@@ -17,9 +18,17 @@ class SkosCodeListNewUriHelper:
...
+    default_uri_style: URIStyle
 
     def _get_identifier_for_document(self) -> str:
-        return f"{self.code_list.metadata.uri_safe_identifier}.csv"
+        extension = (
+            "" if (
+                self.code_list == None
+                or self.code_list.uri_style == URIStyle.WithoutFileExtensions
+                or self.default_uri_style == URIStyle.WithoutFileExtensions
+            ) else ".csv"
```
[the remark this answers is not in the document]
goofballLogic: fixed

csvcubed/csvcubed/writers/qbwriter.py
```diff
@@ -81,6 +82,8 @@ def __post_init__(self):
             self.csv_file_name,
         )
         self._new_uri_helper = QbCubeNewUriHelper(self.cube)
+        self.extension = (".csv" if self.cube.uri_style != URIStyle.WithoutFileExtensions else "")
```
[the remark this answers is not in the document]
goofballLogic: done

csvcubed/csvcubed/writers/qbwriter.py
```diff
@@ -96,7 +99,7 @@ def write(self, output_folder: Path):
 
         tables = [
             {
-                "url": self.csv_file_name,
+                "url": self._new_uri_helper.get_identifier_for_document(),
```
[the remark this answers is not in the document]
goofballLogic: fixed

csvcubed/csvcubed/writers/qbwriter.py
```diff
@@ -196,7 +199,7 @@ def _get_table_references_needed_for_foreign_keys(self) -> List[dict]:
 
                 tables.append(
                     {
-                        "url": f"{code_list.metadata.uri_safe_identifier}.csv",
+                        "url": f"{code_list.metadata.uri_safe_identifier}{self.extension}",
```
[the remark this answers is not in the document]
goofballLogic: fixed

csvcubed/csvcubed/writers/qbwriter.py
```diff
@@ -228,14 +231,13 @@ def _generate_foreign_keys_for_cube(self) -> List[dict]:
...
                 )
-
                 foreign_keys.append(
                     {
                         "columnReference": csvw_column_name_safe(
                             col.uri_safe_identifier
                         ),
                         "reference": {
-                            "resource": f"{code_list.metadata.uri_safe_identifier}.csv",
+                            "resource": f"{code_list.metadata.uri_safe_identifier}{self.extension}",
```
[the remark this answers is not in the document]
goofballLogic: fixed

csvcubed/csvcubed/writers/urihelpers/skoscodelist.py
```diff
@@ -17,9 +18,24 @@ class SkosCodeListNewUriHelper:
...
+        resolved_uri_style = self._resolve_uri_style()
+        if resolved_uri_style == URIStyle.Standard:
+            return identifier + ".csv"
+        elif resolved_uri_style == URIStyle.WithoutFileExtensions:
+            return identifier
+        else:
+            raise ValueError(f"Unhandled URI Style '{self.code_list.uri_style}'.")
+
+    def _resolve_uri_style(self) -> URIStyle:
+        if not (self.code_list is None or self.code_list.uri_style is None):
```
robons, line 35: I think I'd prefer `if self.code_list is not None and self.code_list.uri_style is not None` or the simplified `if self.code_list and self.code_list.uri_style` - I've been told (and believe) that negatives are harder to understand.

csvcubed/tests/behaviour/steps/qbwriter.py
```diff
@@ -1,7 +1,11 @@
+from lib2to3.pytree import Node
```
robons, line 1: I hope we don't need this import.

csvcubed/tests/unit/writers/test_qbwriter.py
```diff
@@ -432,6 +503,57 @@ def test_default_property_value_uris_single_measure_obs_val():
...
+        ],
+    )
+    qb_writer = QbWriter(cube)
+    with TemporaryDirectory() as temp_dir:
+        temp_dir = Path(temp_dir)
+        qb_writer._output_new_code_list_csvws(temp_dir)
+        graph = Graph()
+        graph.parse(temp_dir / "some-dimension.csv-metadata.json")
+        assert (
+            URIRef(f"file://{temp_dir}/some-dimension.csv#code-list"),
```
robons, line 525: This may well cause some problems on Jenkins if the file/folder path contains chars like `@`, etc. could you use a `publicID` when you parse instead so we can be sure to side-step the problem?

robons (approved):
I'll call it approved if you fix the couple of issues I'd raised. We shouldn't need another review I don't think.

csvcubed/csvcubed/writers/urihelpers/skoscodelist.py
goofballLogic: fixed

PR #396, opened 2022-04-06
Implementing ics via csv+csvw

mikeAdamss:
Note : there is some cruft in the git history, will try and clean it out but concentrate on the mentioned files only please

### What is this?

Looking at the qb integrity constraints as currently applied as SPARQL "tests" ran against outputs from the legacy pipelines (see out implementation [here](https://github.com/GSS-Cogs/gdp-sparql-tests/tree/master/tests/qb)), to:

- Identify which of these can be reused against a pure csv+csvw representation of a dataset.
- Create a spike code implementation of how we can apply said integrity constraints using a performant pure python solution that could be applied in an "any dataset of any size" scenario.
- Identify future steps for how we can satisfy the integrity constraints:
             (a) in the short term (i.e stopgap measures for our mvp rdf structures).
             (b) once the currently missing aspects of the qb structure are added.
- Some examples.

### This spike

Contains two principle things:

#### 1.) Spike Code

As included in this draft PR. I've put it all in a single directory as while the logic is  coupled to csvcubed's current implementation, that's not necessarily the end scenario. There is a (potential) future conversation about where such logic/code should live so I've tried to keep it distinct.
  
 In terms of viewing, just look at `./constraints/ics.py`. The other code is ideas of how we could do this in a scalable performant way, but `ics.py` is the final product and part you'd extend for better test coverage.
 
 This `ics.py` contains two examples from the 20 standard `qb` integrity constraints. One example of a straight python implementation of a constraint, one an example of a stopgap mechanism we could apply.
 
 Other than these examples, this code is mainly ideas around how to check these constraints performantly and with a firm eye on avoiding resource issues from unknown file sizes.
 
 #### 2.) The Other ICs
 
The attached google sheet contains some thoughts and documentation on how you could apply the other 18 `qb` integrity constraints using the mechanisms outlined in this spike code going forward. https://docs.google.com/spreadsheets/d/1gAnGACHVhUL02RS0CTTES4NaSsUBBBPc9mr59VaO7RU/edit#gid=0

### Summary

I've only looked at the principle `qb` contraints but the same techniques should be applicable across the board. There's work here certainly, but it should be possible to do this.

_Edit: Alex made an interesting point about relying on pandas (its big), the nature of this approach (discrete row reading) should lend itself quite well to a non pandas approach should we want to go that way, it'll be more complex but not hideously so. If this ends up as a separate tool I'd certainly consider going that route._

### DIFF
--- a/csvcubed/csvcubed/constraints/columns.py
+++ b/csvcubed/csvcubed/constraints/columns.py
@@ -1,0 +1,219 @@
+from dataclasses import dataclass
+from enum import Enum
+import json
+from typing import Dict, List, Optional
+
+
+class ColType(Enum):
+    DIMENSION = "dimension"
+    MEASURE = "measure"
+    ATTRIBUTE = "attribute"
+    MEASURE_TYPE = "measure type"
+
+
+@dataclass
+class ColumnRef:
+    """
+    A class to hold the information from the [tables][tableSchema] dictionary of the csvw
+    defining a column, along with additional information that can inferred from that.
+    """
+
+    csvw_as_dict: Optional[dict]  # pointer
+    initial_column_dict: dict
+
+    # (possible) existing values
+    name: Optional[str]
+    title: Optional[str]
+    required: Optional[str] = None
+    propertyUrl: Optional[str] = None
+    valueUrl: Optional[str] = None
+
+    # values we (possibly) are inferring
+    implicit_component_postfix: Optional[str] = None
+    implicit_url_differentiator: Optional[str] = None
+    implied_type: Optional[ColType] = None
+
+    def serialised(self) -> str:
+        """
+        Simple json serialiser for debugging
+        """
+        jsonified = {
+            "from_csvw_column_definition": {
+                "name": self.name,
+                "title": self.title,
+                "required": self.required,
+                "propertyUrl": self.propertyUrl,
+                "valueUrl": self.valueUrl,
+            },
+            "implied_extrapolated_values": {
+                "implicit_component_postfix": self.implicit_component_postfix,
+                "implicit_url_differentiator": self.implicit_url_differentiator,
+                "implied_type": self.implied_type.value,
+            },
+        }
+
+        return json.dumps(jsonified, indent=2)
+
+    def populate(self):
+        """
+        Populates fields we can infer but don't have to begin with
+        """
+        self.set_implied_type()
+        self.set_implicit_component_postfix()
+        self.set_impicit_url_differentiator()
+
+    # TODO - consider case of not having a propertyUrl
+    def set_implied_type(self):
+        """
+        From the information we have, "guess" the type of column based on the
+        information provided within a given columns propertyUrl
+        """
+        if not self.propertyUrl:
+            return  # TODO - consider ramifications
+
+        matches = []
+
+        if any(["/dimension/" in self.propertyUrl, "/dimension#" in self.propertyUrl]):
+            matches.append(ColType.DIMENSION)
+
+        elif any(
+            [
+                "/attribute/" in self.propertyUrl,
+                "#attribute/" in self.propertyUrl,
+                "/attribute#" in self.propertyUrl,
+            ]
+        ):
+            matches.append(ColType.ATTRIBUTE)
+
+        elif "/measure/" in self.propertyUrl:
+            matches.append(ColType.MEASURE)
+
+        elif "/cube#measureType" in self.propertyUrl:
+            matches.append(ColType.MEASURE_TYPE)
+
+        else:
+            raise ValueError(
+                f"Unable to identify column type from propertyUrl {self.propertyUrl}"
+            )
+
+        assert len(matches) == 1, (
+            f"Logic error. {self.propertyUrl} is matching"
+            "more than one implied column type."
+        )
+        self.implied_type = matches[0]
+
+    def set_implicit_component_postfix(self):
+        """
+        The end part of the implied component identifier being used by csvcubed,
+
+        example:
+        uk-services-trade-by-business-characteristics.csv#component/period
+        from:
+        <SOME-ROOT-DOMAIN/uk-services-trade-by-business-characteristics.csv#component/period>
+        """
+        url = self.csvw_as_dict["tables"][0]["url"]
+        self.implicit_component_postfix = (
+            f'{url}#component/{self.name.replace("_", "-")}'
+        )
+
+    def set_impicit_url_differentiator(self):
+        """
+        By "impicit_url_differentiator" I mean the bit on the end of urls that
+        renders them unique within a given namespace.
+
+        Examples:
+        refPeriod        : http://purl.org/linked-data/sdmx/2009/dimension#refPeriod
+        flow-directions  : http://gss-data.org.uk/def/trade/property/dimension/flow-directions
+        """
+
+        if "propertyUrl" not in self.initial_column_dict:
+            return  # TODO - consider ramifications
+
+        property_url_value = self.initial_column_dict["propertyUrl"]
+        implicit_identifier_index = max(
+            property_url_value.rfind("/"), property_url_value.rfind("#")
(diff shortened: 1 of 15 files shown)


csvcubed/csvcubed/constraints/virtualiser.py
```diff
@@ -0,0 +1,196 @@
...
+        rowcount of MAXIMUM_ROW_SUBSET
+        """
+
+        first_iteration = True
+
+        for start, end in self._subset_index_yieder():
+            # Keep track of the headers from the first iteraton and reuse them
+            # (names) for later dataframe slices that don't have a header row
+            if first_iteration:
+                one_col_df: pd.DataFrame = pd.read_csv(
```
robons, line 137: Interesting approach to iterating over a large dataframe.

I do wonder whether it would be better performance-wise to read the CSV into a sqlite database which we could then index on-disk. We wouldn't need to attempt to keep things in memory and our queries could be executed against the on-disk indexes database table. It might be nice to compare the performance of these two different approaches.

csvcubed/csvcubed/constraints/virtualiser.py
```diff
@@ -0,0 +1,196 @@
...
+                    self.csv_path, usecols=[column_label], skiprows=start, nrows=end
+                )
+                headers = one_col_df.columns.values
+                first_iteration = False
+            else:
+                one_col_df: pd.DataFrame = pd.read_csv(
+                    self.csv_path,
+                    names=headers,
+                    usecols=[column_label],
```
robons, line 146: Ah, so this only reads one column at a time from the CSV?

csvcubed/csvcubed/constraints/virtualiser.py
```diff
@@ -0,0 +1,196 @@
...
+                    skiprows=start,
+                    nrows=end,
+                )
+
+            one_col_df.fillna("", inplace=True)
+            yield one_col_df
+
+    def _column_value_yielder(self, column_label: str) -> str:
```
robons, line 154: Shouldn't the type be `Iterable[str]` here?

csvcubed/csvcubed/constraints/virtualiser.py
```diff
@@ -0,0 +1,196 @@
...
+            for cell_value in one_col_df[column_label]:
+                yield cell_value
+
+    def _unique_column_value_yielder(self, column_label: str) -> str:
+        """
+        A generator that yields one value taken from a single csv column
+        """
+
+        # If we've done this before, use the cached values
+        if column_label in self.unique_column_value_cache:
```
robons, line 173: Might be worth using the [lru_cache decorator](https://docs.python.org/3/library/functools.html#functools.lru_cache) for things like this.

csvcubed/csvcubed/constraints/virtualiser.py
```diff
@@ -0,0 +1,196 @@
...
+
+        else:
+            yielded = []
+
+            one_col_df: pd.DataFrame
+            for one_col_df in self._column_subset_yielder(column_label):
+
+                # new values
+                values_we_havnt_returned = [
+                    x for x in one_col_df[column_label].unique() if x not in yielded
```
robons, line 185: This would be more performant for large columns if `yielded` were a set rather than a list.

robons (commented):
Thanks for the thorough work on this spike.

It certainly makes it clearer to me that we need to prioritise some work on defining dependencies between cubes, their code-lists and any external resources they happen to be re-using. Once we've got that sorted, we'll be able to write equivalent tests to cover a lot of the SPARQL tests we currently have.

I think it's probably worthwhile setting up a new spike on the back of this one to investigate how performantly we can get a CSV inserted into a suitable indexed sqlite database using python - and how easy it is to switch from in-memory to on-disk storage when datasets are too big to hold everything in memory.

csvcubed/csvcubed/constraints/virtualiser.py
mikeAdamss: yeah that was the idea I was playing with, only ever be reading in x rows of a given column at a time as a means to minimise/control memory footprint. I was worried the overhead of just x rows would fluxuate based on number of columns.

mikeAdamss:
cheers for looking at this, I know the code is a messy scramble of ideas. Good thoughts and a great point on leaning on the sparql queries more, ty 👍 

I've got a few questions on things I need to clarify, just want to capture them while I'm reading

- do we want this to be internal to csvcubed or its own thing? Installing pandas and a database is quite heavy duty to read rows, but obviously much less of a consideration if we're doing it anyway in csvcubed.

I guess I'm thinking about one of the questions about whether we need to use pandas in csvcubed, no horse in that race, but it plays into this a bit. I was if anything thinking I'd be taking the pandas reliance out. I didn't do anything here that actually needs a dataframe, at least in so far as i went (though I didn't get to deduplication, that feels like the obvious challenging bit).

- are we only concerned with csvcubed produced csvw's? It's still a bit muddy for me, so say csvcubed can never do X, it feels odd to validate that nobody has done X. It feels related to initial questions as if its general csvw checking - its its own thing - dependency considerations come into play, if that makes sense,

csvcubed/csvcubed/constraints/columns.py
```diff
@@ -0,0 +1,219 @@
...
+        """
+        From the information we have, "guess" the type of column based on the
+        information provided within a given columns propertyUrl
+        """
+        if not self.propertyUrl:
+            return  # TODO - consider ramifications
+
+        matches = []
+
+        if any(["/dimension/" in self.propertyUrl, "/dimension#" in self.propertyUrl]):
```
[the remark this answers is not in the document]
robons: > cheers for looking at this, I know the code is a messy scramble of ideas. Good thoughts and a great point on leaning on the sparql queries more, ty 👍
> 
> I've got a few questions on things I need to clarify, just want to capture them while I'm reading
> 
>     * do we want this to be internal to csvcubed or its own thing? Installing pandas and a database is quite heavy duty to read rows, but obviously much less of a consideration if we're doing it anyway in csvcubed.
> 

So ideally it would be part of csvcubed, I imagine a command called `test` which could take a CSV-W input and run all of the tests we have against it (along with the validations we do when the user builds a cube).

I think pandas has been fine for what we've been working with so far but I think the complexity of the SPARQL queries combined with potentially large file sizes means using sqlite for this would probably be a better route to go down. I think the sqlite dependency is only a few mb tbh and is [installed by default with python](https://www.tutorialspoint.com/sqlite/sqlite_python.htm). 

> 
> I guess I'm thinking about one of the questions about whether we need to use pandas in csvcubed, no horse in that race, but it plays into this a bit. I was if anything thinking I'd be taking the pandas reliance out. I didn't do anything here that actually needs a dataframe, at least in so far as i went (though I didn't get to deduplication, that feels like the obvious challenging bit).
> 

I think finding duplicates should be quite simple with an appropriate sqlite table and query. I have certainly written Microsoft-SQL queries that highlight duplicate rows in the past so I'm fairly confident a similar query can be written in sqlite; you just have to group by the dimension + measure columns and return rows `HAVING COUNT(obs_val)  > 1`.

>     * are we only concerned with csvcubed produced csvw's? 

Yes. This is one of the muddy things with our project at the moment that people are talking about a scale of CSV-W maturity. Honestly, I think our primary purpose now is advocating for csvcubed or nothing. There's basically no way we can cope with all of the different ways people could possibly build CSV-Ws to represent the same information, so I'm planning on standardising on a format akin to what we're producing now so we can more easily write tools which extract the information reliably.

> It's still a bit muddy for me, so say csvcubed can never do X, it feels odd to validate that nobody has done X. 

This is a bit of a muddy one, yes. The problem is that users can still manually edit the CSVs and CSV-Ws and cause problem X to happen, so I think it's a good idea (if lower priority) to be able to highlight those problems.

> It feels related to initial question as if its general csvw checking - its its own thing - dependency considerations come into play, if that makes sense,

PR #403, opened 2022-04-07
Fixing broken template lookup functionality.

robons:
These fixes come from the release candidate branches/tags which we had to make in a hurry yesterday to ensure we released working code for the demo.


canwaf (approved):
Looks good, the number of times I added `cubeconfig` to that `TEMPLATE_BASE_URL` in the last 24 hours means I will inevitably think about this in 20 years unprompted.

Issue #389, opened 2022-04-05
Bug reading in CSV-W code-list from file using the mixed-code-list syntax


robons:
In [skoscodelistreader.py#L55](https://github.com/GSS-Cogs/csvcubed/blob/2150af168303749877f30b519f79560624851675/csvcubed/csvcubed/readers/skoscodelistreader.py#L55) we try to parse some information from code-lists which are defined in CSV-Ws on disk. 

Unfortunately when passing through a mixed-code-list with concepts from different existing code-lists we end up raising the following validation error:

![image.png](https://images.zenhubusercontent.com/60d35dd0922789587fc8922d/8409c70a-25ae-4612-b33e-1b0ddbea08d1)

[example of CSV-W code-list for which this function raises the exception](https://github.com/GSS-Cogs/family-trade/blob/master/datasets/ONS-International-trade-in-services-by-subnational-areas-of-the-UK/codelists/location.csv-metadata.json).

- [x] Rework the code in skoscodelistreader.py so that we can successfully load the example CSV-W provided.
- [x] Write a behaviour test to ensure the whole build process can cope with one of these kind of CSV-Ws.

(This bug comes from @Shannon95 and @santhosh-thangavel)

PR #420, opened 2022-04-11
Fix to ensure that legacy composite code lists can be used as code-lists

robons:
Fixes issue #389.


csvcubed/csvcubed/readers/skoscodelistreader.py
```diff
@@ -52,10 +58,18 @@ def extract_code_list_concept_scheme_info(
             + f"Expected 1, found {len(variables_in_about_url)}"
         )
 
-    if "notation" not in variables_in_about_url:
-        raise ValueError(
-            "Unexpected variable found in aboutUrl template. Expected 'notation', "
-            + f"found '{variables_in_about_url.pop()}'"
+    variable_name_in_about_url = first(variables_in_about_url)
+    assert variable_name_in_about_url is not None
```
GDonRanasinghe, line 62: I suggest we throw an exception here instead of asserting.

csvcubed/csvcubed/readers/skoscodelistreader.py
```diff
@@ -52,10 +58,18 @@ def extract_code_list_concept_scheme_info(
...
+
+    if variable_name_in_about_url != "notation":
+        _logger.debug(
+            'Replacing variable name "%s" in URI template with "notation" for consistency.',
+            variable_name_in_about_url,
+        )
+        about_url = re.sub(
+            "\\{(.?)" + re.escape(variable_name_in_about_url) + "\\}",
+            "{\\1notation}",
```
GDonRanasinghe, line 71: Do we need to guarantee the uniqueness of the variable name? If so, we can use uuid() here.

GDonRanasinghe (commented):
@robons Added a couple of comments. Please let me know what you think about them.

GDonRanasinghe (approved):

csvcubed/csvcubed/readers/skoscodelistreader.py
robons: I'd normally agree with you, but that assert statement will always be false and is only there to make it clear to pyright that the type is `str` rather than `Optional[str]`.

csvcubed/csvcubed/readers/skoscodelistreader.py
robons: No, we don't want it to be unique - in fact we want it to be consistent. It gets replaced in the caller by a suitable value.

PR #422, opened 2022-04-12
Update tests to write to a TemporaryDirectory.

Giraffe-Technology:


csvcubed/tests/behaviour/steps/cubes.py
```diff
@@ -58,9 +59,15 @@ def step_impl(context):
     context.add_cleanup(lambda: mocker.stop())
 
     with vcr.use_cassette(str(_cassettes_dir / "cube-created.yaml")):
-        cube, errors = cli_build(data_file, config_file)
-        context.cube = cube
-        context.errors = errors
+        with TemporaryDirectory() as temp_dir_path:
```
robons, line 62: Whilst this would be fine for the unit test approach, with the behaviour tests we have a common temporary directory in which we stick all of the outputs - this means the next step can access the resources created in previous steps.

There's a function somewhere to get the temp dir given the context. If you could use that instead of creating a new temporary directory in this scope, then this would be fine.

robons: The unit tests look good though.

Giraffe-Technology: Behaviour test updated, also had a little clean and format.

PR #424, opened 2022-04-12
392 refactor inspect metadata printer class

GDonRanasinghe:
Refactored the MetadataPrinter dataclass so it uses the __post_init()__ methods.


canwaf (approved):
Easily approved.

canwaf (approved):

Issue #360, opened 2022-03-04
CSV-W output data types not being respected


canwaf:
csvcubed doesn't output observation columns respecting the data_type as set in the info.json (or probably qube-config.json). 

For example, [Sweden at Eurovision](https://github.com/canwaf/csvcubed-docker/tree/master/sweden_at_eurovision) has set the observation column to be int, and the output of the csv-column when `infojson2csvqb` is run comes out as a float (e.g. `1` gets serialised as `1.0`); despite the output of corresponding `metadata.json` file saying the column is an int.

This bug is confirmed in the observation column; however likely also present in literal attribute columns.

You'll need to find where the `data.to_csv(... index=false, ...)` to insert a data_type alignment function before serialising the csv to disk.

This means that csvlint will fail. Example error:

```bash
# docker run -it --rm -v $(pwd):/workspace -w /workspace gsscogs/csvlint csvlint -s sweden-at-eurovision.csv-metadata.json
.!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!.!!!!!!!!!.!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!.!!!!!!!!!.!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!.!
/workspace/sweden-at-eurovision.csv is INVALID
1. invalid_int. Row: 2,7. 4.0
2. invalid_int. Row: 3,7. 9.0
3. invalid_int. Row: 4,7. 10.0
4. invalid_int. Row: 5,7. 14.0
5. invalid_int. Row: 6,7. 7.0
6. invalid_int. Row: 7,7. 13.0
7. invalid_int. Row: 8,7. 10.0
8. invalid_int. Row: 9,7. 2.0
9. invalid_int. Row: 10,7. 8.0
10. invalid_int. Row: 11,7. 5.0
11. invalid_int. Row: 12,7. 9.0
```

PR #428, opened 2022-04-13
Issue #360 - Fixing test problems on windows.

robons:
Most of the issues are naturally to do with how we handle paths. We need to start making an extra effort to ensure we're using pathlib in a way which minimises the impact across different platforms.

We should also find a way to add windows platform tests to our release process - perhaps we can use the VM in GCP for now as a manual process but look to move to github actions (which I believe support running stuff on windows) at a later point.

I have *not* been able to run docker tests yet. But it's important to get these fixes in ASAP regardless.


GDonRanasinghe (approved):
This PR looks good to me. No further comments. 

The pipeline seems to have failed for some reason. I've restarted it.

PR #432, opened 2022-04-14
Fixes added to the dockerfile

GDonRanasinghe:
Since we have `apt-get update` and `apt-get install` in separate `RUN` statements inside the Dockerfile, the latter command was failing when building the container (without cache).

Docker best practices recommend combining the two commands into a single `RUN` statement since "Using apt-get update alone in a RUN statement causes caching issues and subsequent apt-get install instructions fail." This PR contains this update.

See https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run for reference.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
 FROM python:3.9
 
-RUN apt-get update
-
-# install docker stuff
-RUN apt-get install -y \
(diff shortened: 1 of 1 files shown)


canwaf (approved):
All good.

Issue #415, opened 2022-04-08
Fix attribute column type in qube-config schema


canwaf:
Schema has a few issues.

```json
{
  "Marker": {
     "from_template": "observation-status"
  }
}
```

- [ ] `"type": "attribute"` is selected, VSC it highlights it as invalid type; however it does recommend the correct keys for attributes
- [ ] This impacts `"from_template": "observation-status"` as it is not recommended if the user doesn't define the `type` as above

The schema file is [here](https://github.com/GSS-Cogs/csvcubed/blob/main/csvcubed/csvcubed/schema/cube-config/v1_0/schema.json).

Issue #414, opened 2022-04-08
Change qube-config schema to support both date and datetime for datetime fields


canwaf:
All datetime fields should accept dates as well. This requires schema changes.

- [ ] issued
- [ ] modified
- [ ] anything else you find that has dates

https://github.com/GSS-Cogs/csvcubed/blob/main/csvcubed/csvcubed/schema/cube-config/v1_0/schema.json

- [ ] make the above schema changes
- [ ] (Behaviour) Test that csvcubed will load and output a CSV-W using just dates in all fields which are currently date-time.
  - This could be a pydantic validation error *or* it might just be the code we're using to read the ISO8601 datetimes in.

N.B. @GDonRanasinghe suggests that CSV-Ws which contain dates rather than date-times *may* break the `inspect` command. As part of this task we need to write a behaviour test to ensure that the inspect command will work as anticipated.

PR #450, opened 2022-04-22
JSON Schema fixes + Supporting parsing dates/date-times from qube-config.json 

robons:
Satisfies issues #414, #415.

image

image

image

### DIFF
--- a/csvcubed/csvcubed/schema/cube-config/v1_0/schema.json
+++ b/csvcubed/csvcubed/schema/cube-config/v1_0/schema.json
@@ -26,17 +26,20 @@
         "keywords": {"$ref": "#/definitions/v1.0/keywordTypes"},
         "dataset_issued": {
             "description": "Date time that the data set was initially published in ISO 8601 format, e.g. 2022-03-31 or 2022-03-31T12:54:30Z",
-            "type": "string",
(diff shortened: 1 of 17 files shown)


csvcubed/csvcubed/schema/cube-config/v1_0/schema.json
```diff
@@ -26,25 +26,29 @@
...
             "description": "Date time that the data set was initially published in ISO 8601 format, e.g. 2022-03-31 or 2022-03-31T12:54:30Z",
-            "type": "string",
-            "format": "date-time"
+            "type": "string"
         },
         "dataset_modified": {
             "description": "Date time that the data set was last modified published in ISO 8601 format, e.g. 2022-03-31 or 2022-03-31T12:54:30Z",
-            "type": "string",
-            "format": "date-time"
+            "type": "string"
```
robons, line 33: You'll notice I've left the date-time fields as just "string". We parse the values as ISO-8601 date/date-times, however the JSON schema specification says that the `date-time` format requires [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339) formatted date-times which differ slightly.

canwaf (approved):

PR #451, opened 2022-04-23
Improved logic when displaying number of observations

GDonRanasinghe:
Improved logic when displaying number of observations.


robons (approved):

robons:
Just make sure you fix whatever has caused the build to fail.

Issue #477, opened 2022-05-09
Date time code lists not being built properly


robons:
Date time code-lists should be built so that they define the periods used in the dataset, but the code list should link those local concepts back to the original reference.data.gov.uk identifiers for the period. This is working correctly in gss-utils/infojson2csvqb, but isn't working in csvcubed at the moment.

PR #478, opened 2022-05-09
Correcting date time code list serialisation

robons:
The PR addresses a couple of bugs:

1. `<column_name>` substitution in column templates was generating invalid URI templates which didn't correctly reference the column.

2. The code which generated local date-time code lists was using the wrong property to make the decision. No testing existed to validate this.
 
Addresses issue #477

### DIFF
--- a/csvcubed/csvcubed/readers/cubeconfig/v1_0/columnschema.py
+++ b/csvcubed/csvcubed/readers/cubeconfig/v1_0/columnschema.py
@@ -15,7 +15,7 @@ import uritemplate
 from csvcubedmodels.dataclassbase import DataClassBase
 
 from csvcubed.inputs import pandas_input_to_columnar_optional_str
-from csvcubed.models.cube import CatalogMetadata
+from csvcubed.models.cube import CatalogMetadata, NewQbConcept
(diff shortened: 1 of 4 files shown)


csvcubed/csvcubed/readers/cubeconfig/v1_0/columnschema.py
```diff
@@ -98,35 +100,40 @@ def _get_code_list(self, new_dimension: NewQbDimension) -> Optional[QbCodeList]:
             elif (
                 new_dimension.parent_dimension_uri
                 == "http://purl.org/linked-data/sdmx/2009/dimension#refPeriod"
-                and self.definition_uri is not None
-                and self.definition_uri.lower().startswith(
+                and self.cell_uri_template is not None
+                and self.cell_uri_template.lower().startswith(
                     "http://reference.data.gov.uk/id/"
```
GDonRanasinghe, line 105: I suggest we define urls like these in a constants or a config file. In this way, if we want to change these, we can easily do so.

GDonRanasinghe (approved):
The PR looks good and ddded a minor suggestion.

PR #499, opened 2022-05-26
validate against NewQbUnit identifiers, not labels

mikeAdamss:
Literally Robs example code, I can't think of any additional test coverage required beyond what we already have.

Happy to take a steer on that, finding this one extremely opaque.

Note: draft for now as I want to understand why the tests were passing previously, but comments still welcome.

Edit: tests were passing previously as we were using the uri template `"{+unit}"` which meant that _coincidentally_ the `unique_expanded_uris` derived from the template + values (made uri_safe) matched the `unit.uri_safe_identifier`'s for `NewQbUnit`'s.


mikeAdamss:
pushed a typo'd commit message. It's only one commit so creating a new PR to correct.

PR #500, opened 2022-05-26
Issue 490/defining units bug

mikeAdamss:
Code from session with Rob, only change is I had to remove the comprehension approach to satisfy PyRight.

Also includes a change to use `columntemplate` over `uri_safe_identifier` for comparisson. This was another bug (obscured by the first bug), tests were only previously passing as we were using the uri template "{+unit}" which meant that _coincidentally_ the `unique_expanded_uris` derived from the template + values (made uri_safe) matched the `unit.uri_safe_identifier`'s for `NewQbUnit`'s which we were using in error.

I don't think it warrants new tests now the existing tests are working, but happy to take a steer, found this very opaque.

### DIFF
--- a/csvcubed/csvcubed/models/cube/qb/components/unitscolumn.py
+++ b/csvcubed/csvcubed/models/cube/qb/components/unitscolumn.py
@@ -88,7 +88,9 @@ class QbMultiUnits(QbColumnStructuralDefinition):
                     map_label_to_new_uri_value.update({u.label: u.uri_safe_identifier})
 
             if map_label_to_new_uri_value:
-                unique_values = {map_label_to_new_uri_value.get(v, v) for v in unique_values}
+                unique_values = {
+                    map_label_to_new_uri_value.get(v, v) for v in unique_values
+                }
 
             unique_expanded_uris = {
                 uritemplate.expand(csv_column_uri_template, {csvw_column_name: s})
@@ -100,7 +102,10 @@ class QbMultiUnits(QbColumnStructuralDefinition):
                     expected_uris.add(unit.unit_uri)
                 elif isinstance(unit, NewQbUnit):
                     expected_uris.add(
-                        uritemplate.expand(csv_column_uri_template, {csvw_column_name: unit.uri_safe_identifier})
+                        uritemplate.expand(
+                            csv_column_uri_template,
+                            {csvw_column_name: unit.uri_safe_identifier},
+                        )
                     )
                 else:
                     raise Exception(f"Unhandled unit type {type(unit)}")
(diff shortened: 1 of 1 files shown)


scheung38:
Could team please remove my email please as I am still bombarded with lots of emails thank you


Sebastian

mikeAdamss:
> 

Hey Seb, you were removed ages ago. 

I believe in this instance you're _watching_ this repository. I would suggest you unwatch and/or update your notifications. I'm pretty sure this is your github settings more than anything I/we have control over.

csvcubed/csvcubed/models/cube/qb/components/unitscolumn.py
```diff
@@ -82,7 +80,18 @@ def validate_data(
         column_csv_title: str,
     ) -> List[ValidationError]:
         if len(self.units) > 0:
-            unique_values = {uri_safe(v) for v in set(data.unique())}
+            unique_values = set(data.unique())
+
+            map_label_to_new_uri_value = {}
+            for u in self.units:
+                if isinstance(u, NewQbUnit):
+                    map_label_to_new_uri_value.update({u.label: u.uri_safe_identifier})
```
robons, line 88: Why did you have to remove the set comprehension?

csvcubed/tests/unit/models/test_unit.py
```diff
@@ -65,6 +65,28 @@ def test_known_existing_units_defined():
...
+        "Unit",
+        QbMultiUnits(
+            [
+                ExistingQbUnit("http://example.org/units/NUM"),
+                ExistingQbUnit("http://example.org/units/PERCENT"),
+            ]
+        ),
+    )
+
+    unit_dimension = unit_column.structural_definition
```
robons, line 82: It's not a dimension, it's just some kind of DSD definition which lists the units used in the dataset.

And I'd suggest the following approach would be a bit simpler and closer to the behaviour I'd expect to happen:

```python
    data = pd.DataFrame({"Unit": ["NUM", "PERCENT"]})
    unit_column = QbColumn(
        "Unit",
        QbMultiUnits(
            [
                ExistingQbUnit("http://example.org/units/NUM"),
                ExistingQbUnit("http://example.org/units/PERCENT"),
            ]
        ),
        csv_column_uri_template="http://example.org/units/{+unit}"
    )
    
    errors = unit_column.validate_data(data["Unit"])

    assert_num_validation_errors(errors, 0)
```

robons (approved):
Looks good to me, although I've made a suggestion on how the unit test could be made to align a bit better with the way the code ends up being run.

csvcubed/csvcubed/models/cube/qb/components/unitscolumn.py
mikeAdamss: pydantic, it doesent logically infer the type from an all(isinstance) conditional , leads to getting errors for QbUnit not having .uri_safe identifer.

works where you use a single condition isinstance.

i dislike the necessity tbh, but i didnt want to to alter the types.

Issue #457, opened 2022-04-25
Update csvcubed project


GDonRanasinghe:
- [x] Project Structuring
    - [x] Project should only contain relevant packages, source code and tests (i.e. unit and feature tests).
    - [x] Project should contain an individualised dev container and the corresponding docker file.
    - [x] The API docs (i.e. sphinx docs) needs to be removed.
    - [x] The csvcubed project has the `docs/` folder which contains the external user docs. 
    - [x] Project should contain the below project structure. The structure is described in [here](https://drive.google.com/file/d/19eSxZLfTL6uMAFcUs2WHMUpCAKc5u4UM/view?usp=sharing).

**Use existing repo for this task, do not create a new repo for this task**

```
- <project-name>
    - .devcontainer
      - devcontainer.json
    - Dockerfile
    - .gitignore
    - docs
    - src
      - <package-name>
        - __init__.py
        - <rest of the source code>
    - tests
      - test-cases
      - unit
      - features
    - poetry.lock
    - pyproject.toml
    - tox.ini
    - Jenkinsfile
    - README.md
    - LICENSE Apache V2
```

- [x] Project Jenkins Pipeline Changes
   - [x] Project will contain a Jenkins file and it will contain steps relevant for that project only. This means, we need to remove all the irrelevant content from the existing Jenkins file for the particular project, and thus individualise the Jenkins file to the individual project.

## Closing tasks
- [x] Verify there are no more relative imports in pyproject.toml
- [x] Update documentation to say you no longer have to install csvcubed-models explicitly

PR #510, opened 2022-06-02
457 csvcubed project update

GDonRanasinghe:
Satisfies #457 - csvcubed project update after splitting the repo.


GDonRanasinghe:
@robons This is a PR for csvcubed repo update. I have marked this as a draft PR for now due to the following consideration:

The approach we have for maintaining only the relevant history only works if we are pushing the code to a new repo. However in the csvcubed project, we aim to maintain the relevant history within the existing csvcubed project - we want to filter out the history related to csvcubed-pmd, csvcubed-devtools etc from the csvcubed repo. To this end, the git filter-repo tool is not suitable as far as I can see. So I have the following questions:

1. Is it correct when I said we want to filter out the history related to now irrelevant files in the csvcubed repo?
2. If the answer to above is Yes, are we open to create a new csvcubed repo?

pyproject.toml
```diff
@@ -48,4 +43,4 @@ exclude = [
 pythonVersion = "3.9"
 
 [tool.poetry.scripts]
-csvcubed = 'csvcubed.cli.entrypoint:entry_point'
\ No newline at end of file
+csvcubed = 'src.csvcubed.cli.entrypoint:entry_point'
```
robons, line 46: so we really need `src.` here? I thought setting the `from = "src"` above would be enough for it to realise that's where the entry points are.

robons:
We'll need to update the [csv-cubed/shacl/v1.0](https://purl.archive.org/purl/csv-cubed/shacl/v1.0) PURL when this is released. Before this is merged in, can you make sure to record this action in the following release issue?

robons:
> @robons This is a PR for csvcubed repo update. I have completed all the tasks in this ticket including updates to unit and behave tests, expect maintaining history due to the following consideration:
> 
> The approach we have for maintaining only the relevant history only works if we are pushing the code to a new repo. However in the csvcubed project, we aim to maintain the relevant history within the existing csvcubed project - we want to filter out the history related to csvcubed-pmd, csvcubed-devtools etc from the csvcubed repo. To this end, the git filter-repo tool is not suitable as far as I can see. So I have the following questions:
> 
>     1. Is it correct when I said we want to filter out the history related to now irrelevant files in the csvcubed repo?

It isn't the end of the world if the history remains in there. But I suspect that we can remove the extraneous history, it'll just mean us focusing on the `main` branch and doing a force push once the old history has been removed (this should only happen after the PR has been merged and we take care with the operation).

>     2. If the answer to above is Yes, are we open to create a new csvcubed repo?

I don't want to create a new repo and as discussed above, I think we can get away without doing that.


------

Also, note that the build is failing so I'd take a look into that.

GDonRanasinghe:
> We'll need to update the [csv-cubed/shacl/v1.0](https://purl.archive.org/purl/csv-cubed/shacl/v1.0) PURL when this is released. Before this is merged in, can you make sure to record this action in the following release issue?
> 
> There are probably a large number of other PURLs which will need to be updated at the point that this is merged into main. Can you make sure to provide appropriate guidance on which ones need updating?
> 
> Also, this PR is likely to move the `templates` folder - meaning that existing versions of csvcubed out in the wild will break when this is merged into main. I haven't checked this but can you make sure the templates folder remains where it is?

PURL guidance is attached here:
[purl_updates.csv](https://github.com/GSS-Cogs/csvcubed/files/8844474/purl_updates.csv)

GDonRanasinghe:
> Also, this PR is likely to move the `templates` folder - meaning that existing versions of csvcubed out in the wild will break when this is merged into main. I haven't checked this but can you make sure the templates folder remains where it is?

The templates folder now remains where it was originally.

tests/behaviour/steps/cli.py
```diff
@@ -17,7 +17,6 @@ def step_impl(context, arguments: str):
     dirs = AppDirs("csvcubed-cli", "csvcubed")
     context.csvcubed_log_location = Path(dirs.user_log_dir)
 
-
```
robons, line 20: Hardly the worst problem, but the black formatter should have insisted you keep this line. It likes 2 lines between functions.

robons (dismissed):
Looks good.

Just to let you know that in preparation for this to be merged into master, I've locked the affected PURL URLs at the current commit. This way when this is merged in, no URLs will break and hopefully you and @MuazzamChaud can work together on updating the relevant redirects.

image

N.B. The `/v1` URIs in your CSV file don't need to be updated since they point to other PURL urls.

image

tests/behaviour/steps/cli.py
GDonRanasinghe: You are correct - I had to reset my black formatter and the formatting works as expected now.

robons (approved):
Looks good to go, thanks!

Issue #511, opened 2022-06-06
Define dataset-local inter-CSV-W dependencies


robons:
When attempting to pull a CSV-W from a web address, `pmdutils csvw pull` can pull the schema of the code-list tables, but cannot pull the JSON-LD metadata describing the code lists since said metadata is stored in files which are not referenced from the dataset's JSON-LD document.

In order to continue to avoid having to stick the code list's metadata directly inside the data set's JSON-LD file (so we can encourage a modular structure), we need to be able to define RDF references between the separate CSV-Ws we generate.

As part of this work we need to:

- [x] Ensure that csvcubed defines sensible dependencies between the dataset CSV-W JSON-LD and the code-list dependencies it's consuming.
- [x] Ensure that the dependency definitions are suitably modified when uploaded to PMD so that they continue to make sense even when there are no relative files to speak of.
- [x] update the `pmdutils csvw pull` command so that it can read the RDF dependencies and pull all required files.
- [x] update the `csvcubed inspect` command so that it can read the RDF dependencies and load all of the required metadata.

N.B. We're not considering versioning to be part of this work at the moment.

Base this work on the [void:dataset dependency work](https://gist.github.com/robons/ee063b662dab46dc100848dc3ceb359d) done previously by @robons.

PR #516, opened 2022-06-08
Defining dependencies between CSV-W metadata files.

robons:
Satisfies issue #511.

Dependencies are defined using the VOID vocabulary, e.g.

```ttl
@prefix void: <http://rdfs.org/ns/void#>.

<file:/tmp/a-qube-with-dependencies.csv#dependency/a-code-list> a void:Dataset;
  void:dataDump <a-code-list.csv-metadata.json>;
  void:uriSpace "a-code-list.csv#".

<file:/tmp/a-qube-with-dependencies.csv#dependency/d-code-list> a void:Dataset;
  void:dataDump <d-code-list.csv-metadata.json>;
  void:uriSpace "d-code-list.csv#".
```

I've reworked the inspect command functionality so it now loads these RDF dependencies - this means that we can finally see the code list's label when inspecting a cube (so long as the code list is defined locally with the dataset).

image

Note that when a CSV-W metadata file is a depedency, we only load the JSON-LD document and currently don't attempt to convert the whole CSV-W to RDF.


src/csvcubed/cli/inspect/metadataprocessor.py
```diff
@@ -99,14 +121,110 @@ def load_json_ld_to_rdflib_graph(self) -> Graph:
...
+                "Skipping dependency '%s' as it has already been loaded.",
+                dependency.data_dump,
+            )
+            continue
+
+        _logger.debug(
+            "Loading dataset dependency '%s' covering uriSpace '%s' in dataset '%s'",
+            dependency.data_dump,
+            dependency.uri_space,
+            dependency.data_set,
```
GDonRanasinghe, line 200: Do we need to have 2 loops here? Can't we do the functionality in the 2nd loop within the 1st loop?

GDonRanasinghe (approved):
@robons This work looks good to me. I've added a minor comment around for loop if you can have a look?

src/csvcubed/cli/inspect/metadataprocessor.py
robons: Fair point, I've merged the two of them together now.

PR #521, opened 2022-06-09
Updated long urls used in docs for templates to use PURL.

GDonRanasinghe:
Updated long urls used in docs for templates to use PURL.


robons (approved):
Good spot!

Issue #525, opened 2022-06-15
`csvw:aboutUrl` bug when dimension column contains URI-unsafe characters


robons:
When generating a CSV-W which contains URI-unsafe characters in a dimension column, csv2rdf of the output fails due to invalid `aboutUrl`s being generated.

```
Performing CSV2RDF operation with arguments atmospheric-emissions-greenhouse-gases-by-industry-and-gas-1990-2020.ttl
 Exception in thread "main" java.lang.IllegalArgumentException: Illegal character in fragment at index 206: atmospheric-emissions-greenhouse-gases-by-industry-and-gas-1990-2020.csv#obs/1990,http://gss-data.org.uk/data/gss_data/climate-change/ons-atmospheric-emissions-greenhouse-gases-by-industry-and-gas-1990-2020#concept/sic-2007/consumer-expenditure,ghg-total@mass-of-air-emissions-of-carbon-dioxide-equivalent
 	at java.base/java.net.URI.create(URI.java:906)
 	at java.base/java.net.URI.resolve(URI.java:1089)
 	at csv2rdf.util$resolve_uri.invokeStatic(util.clj:115)
 	at csv2rdf.util$resolve_uri.invoke(util.clj:105)
 	at csv2rdf.metadata.uri_template_property$resolve_uri_template_property.invokeStatic(uri_template_property.clj:22)
 	at csv2rdf.metadata.uri_template_property$resolve_uri_template_property.invoke(uri_template_property.clj:17)
 	at csv2rdf.tabular.csv$get_cell_urls.invokeStatic(csv.clj:138)
 	at csv2rdf.tabular.csv$get_cell_urls.invoke(csv.clj:137)
 	at csv2rdf.tabular.csv$annotate_row$fn__7904.invoke(csv.clj:149)
 	at clojure.core$map$fn__5851.invoke(core.clj:2755)
 	at clojure.lang.LazySeq.sval(LazySeq.java:42)
 	at clojure.lang.LazySeq.seq(LazySeq.java:51)
 	at clojure.lang.RT.seq(RT.java:531)
 	at clojure.lang.LazilyPersistentVector.create(LazilyPersistentVector.java:44)
 	at clojure.core$vec.invokeStatic(core.clj:377)
 	at clojure.core$vec.invoke(core.clj:367)
 	at csv2rdf.tabular.csv$annotate_row.invokeStatic(csv.clj:154)
[...]

PR #526, opened 2022-06-15
Issue #525 - Altering serialisation of `aboutUrl` values so that they can cope successfully with invalid URI chars.

robons:
Satisfies issue #525.

### DIFF
--- a/src/csvcubed/definitions.py
+++ b/src/csvcubed/definitions.py
@@ -8,8 +8,10 @@ Currently defines the root project path.
 import os
 from pathlib import Path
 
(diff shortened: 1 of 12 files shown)


src/csvcubed/definitions.py
```diff
@@ -8,8 +8,10 @@
 import os
 from pathlib import Path
 
+APP_ROOT_DIR_PATH = Path(os.path.dirname(os.path.abspath(__file__)))
```
robons, line 11: This change makes sure that it is correct when packaged up into a whl (at which point this file is at the root of the package).

PR #531, opened 2022-06-22
Issue 452/invalid uri template

mikeAdamss:
pr for https://app.zenhub.com/workspaces/features-squad-61b72275ac896c0010a1b00b/issues/gss-cogs/csvcubed/452

will update purl link once/if the approach gets signed off and before any merge.

I've also included a missing vcr fixture that we're generating but haven't checked into source control.

### DIFF
--- a/src/csvcubed/models/cube/cube.py
+++ b/src/csvcubed/models/cube/cube.py
@@ -101,11 +101,13 @@ class Cube(Generic[TMetadata], PydanticModel):
                         self._get_validation_error_for_exception_in_col(column, e)
                     )
 
+        # Check for uri template naming errors
         safe_column_names = [
             csvw_column_name_safe(c.uri_safe_identifier) for c in self.columns
         ]
         for uri_template, name in self._csv_column_uri_templates_to_names():
             if name not in safe_column_names:
+                _logger.debug('Unable to find name %s in %s', name, safe_column_names)
                 errors.append(UriTemplateNameError(safe_column_names, uri_template))
 
         return errors
@@ -113,8 +115,9 @@ class Cube(Generic[TMetadata], PydanticModel):
 
     def _csv_column_uri_templates_to_names(self) -> Tuple[str, str]:
         """
-        Generates tuples of each configured csv column uri template
-        along with the column name specified within it.
+        Generates tuples of any configured and not None csv column
+        uri templates within the cube, along with the column name
+        specified within it.
 
         Example:
         A cube containing just the following csv column uri templates
@@ -128,11 +131,11 @@ class Cube(Generic[TMetadata], PydanticModel):
         - "http://example.com/dimensions/{bar}#things": "bar"
         """
         csv_column_uri_templates = [
-            c.csv_column_uri_template for c in self.columns if c.csv_column_uri_template
+            c.csv_column_uri_template for c in self.columns if hasattr(c, "csv_column_uri_template")
         ]
         template_to_name_map = {
             c: extract_uri_template_variable_name_by_index(c)
-            for c in csv_column_uri_templates
+            for c in csv_column_uri_templates if c
         }
 
         for uri_template, name in template_to_name_map.items():
--- a/src/csvcubed/models/cube/cube.py
+++ b/src/csvcubed/models/cube/cube.py
@@ -5,12 +5,13 @@ Cube
 import logging
 from dataclasses import dataclass, field
 from typing import List, Optional, Set, TypeVar, Generic, Iterable, Tuple
+
 import pandas as pd
+import uritemplate
 
-from csvcubed.models.validationerror import (
-    ValidationError,
-)
-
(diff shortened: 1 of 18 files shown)


src/csvcubed/utils/uri.py
```diff
@@ -112,3 +112,21 @@ def ensure_values_in_lists_looks_like_uris(values: list[str]) -> None:
...
+    
+    index 0: "http://example.com/something/{+foo}" = "foo" 
+    
+    index 0: "http://example.com/something/{bar}" = "bar"
+    
+    index 0: "http://example.com/{one}/segment/{two}#stuff = "one"
+    
+    index 1: "http://example.com/{one}/segment/{two}#stuff = "two"
+    """
+    vars = [x.split("{")[1] for x in uri_template.split("}")[:-1]]
```
robons, line 131: We're already using the [uritemplate](https://uritemplate.readthedocs.io/en/latest/) package for [stuff like this](https://github.com/GSS-Cogs/csvcubed/blob/55905b98a50c452f0ed999cc8f7ea6f84eae3e22/src/csvcubed/readers/skoscodelistreader.py#L54).

src/csvcubed/models/cube/validationerrors.py
```diff
@@ -92,3 +93,25 @@ def get_error_url(cls) -> str:
...
+@dataclass
+class UriTemplateNameError(SpecificValidationError):
+    """
+    An error to inform the user that they the column name they are referring to in a csv uri template
+    does not exist.
+    """
+
+    column_names: List[str]
+    csv_column_uri_template: str
+    column_names_concatenated: Optional[str] = None
```
robons, line 107: Since this is being populated in the `__post_init__` method, ` = field(init=False)` is the best thing to stick here since it makes sure it doesn't appear in the constructor/`__init__` at all.

tests/unit/readers/cubeconfig/v1_0/test_build_friendly_error_msgs.py
```diff
@@ -270,6 +271,33 @@ def test_val_errors_both_unit_types():
     )
 
 
+def test_val_errors_invalid_uri_template():
+    """
+    Test for:-
+        TBD
```
robons, line 277: todo

tests/unit/readers/cubeconfig/v1_0/test_build_friendly_error_msgs.py
```diff
@@ -270,6 +271,33 @@ def test_val_errors_both_unit_types():
...
+    config = Path(_test_case_dir, "invalid_uri_template.json")
+    csv = Path(_test_case_dir, "invalid_uri_template.csv")
+    cube, json_schema_validation_errors, validation_errors = _extract_and_validate_cube(
+        config, csv
+    )
+    _write_errors_to_log(json_schema_validation_errors, validation_errors)
+
+    assert isinstance(cube, Cube)
+    assert isinstance(validation_errors, list)
+    assert len(validation_errors) == 1
```
robons, line 288: I'd use the [`assert_num_validation_errors(...)`](https://github.com/GSS-Cogs/csvcubed/blob/879b3cb7a899a6fcb22745fa55ec59b0106d4e78/tests/unit/test_baseunit.py#L40) function here since it helps debugging what went wrong.

robons (changes requested):
Looks good, but we need to be able to check more than just the first URI template variable in the URI.

src/csvcubed/models/cube/cube.py
```diff
@@ -96,4 +103,42 @@ def _validate_columns(self) -> List[ValidationError]:
...
+        "http://example.com/dimensions/{+foo}"
+        "http://example.com/dimensions/{bar}#things"
+
+        yields:
+        - "http://example.com/dimensions/{+foo}": "foo",
+        then
+        - "http://example.com/dimensions/{bar}#things": "bar"
+        """
+        csv_column_uri_templates = [
+            c.csv_column_uri_template for c in self.columns if hasattr(c, "csv_column_uri_template") # pyright: ignore
```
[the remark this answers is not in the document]
mikeAdamss: yeah I tried that initially, there's a circular import I couldn't resolve. Happy to take a steer on how or an alternatively but I had no luck unpicking it.

Did get me curious though, is this https://github.com/GSS-Cogs/csvcubed/blob/b597f1ee9e978f1fe36cf8ffedade2b31e6fa8ec/src/csvcubed/models/cube/cube.py#L37  actually be a CsVColumn or are we using the parent class to avoid the circular import?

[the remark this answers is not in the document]
robons: The parent class was there since the cube class was potentially going to be used for different types of outputs (i.e. a QbCube or perhaps we could have configured a CmdCube as well?).

Let's have a call on the circular import issue.

Issue #518, opened 2022-06-08
Code List Config `same_as` field not serialised to CSV-W


GDonRanasinghe:
When the user specifies a `same_as` relation between a concept in their new code list and an existing concept defined elsewhere, the information is not serialised into the code list CSV-W.

We need to make sure that the same as relationship is serialised correctly into the CSV-W. This means adding the column to the CSV file as well as adding the `owl:sameAs` relationship to the column definition.

Refactor behave tests in cube feature.

GDonRanasinghe:
@robons I've updated the effort in this ticket from 3 to 5 as this involved an additional task - refactoring behave tests in the cube feature. I have refactored all the tests in the cube feature.

canwaf:
This is an example of it not serialising the same_as provided in a concept scheme. https://github.com/GSS-Cogs/family-climate-change/tree/master/datasets/Forestry-Research-Forestry-Statistics-2022-Carbon

canwaf:
Scoring: Impact is high: the functionality is not accessible and there is no workaround. Probability is moderate: users who want mixed-concept schemes will be impacted, which account for large number of datasets but not the majority.

This makes this bug critical.

PR #538, opened 2022-07-04
518 code list config same as field serialisation

GDonRanasinghe:
This PR includes the implementations for supporting concepts defined elsewhere when defining a code list using code list config.

Satisfied #518


src/csvcubed/writers/skoscodelistwriter.py
```diff
@@ -38,6 +39,15 @@ class SkosCodeListWriter(WriterBase):
     def csv_metadata_file_name(self) -> str:
         return f"{self.csv_file_name}-metadata.json"
 
+    @staticmethod
+    def has_duplicated_qb_concepts(code_list: NewQbCodeList) -> bool:
+        duplicated_qb_concepts: list[DuplicatedQbConcept] = [
```
robons, line 44: I'd replace this with `return any(True for concept in code_list.concepts if isinstance(concept, DuplicatedQbConcept)` This should hopefully avoid actually creating the list.

tests/behaviour/cube.feature
```diff
@@ -257,4 +257,68 @@ Feature: Cube!
...
+        "keywords": ["Test cube"],
+        "dataset_issued": "2022-04-08",
+        "dataset_modified": None,
+        "license_uri": "https://creativecommons.org/licenses/by/4.0/",
+        "public_contact_point_uri": None,
+        "uri_safe_identifier_override": None
+      }
+      """
+    Then The cube Columns should match
+    """
```
robons, line 285: Can we avoid reusing these kinds of tests where we match on JSON representations of objects? They're really brittle and it's hard to figure out what you're actually testing.

Can't you test something clear like the code-list's CSV output has the appropriate column defined?

robons (changes requested):
Looks sensible, apart from the behaviour tests which aren't clear what they're trying to test.

tests/behaviour/cube.feature
GDonRanasinghe: @robons I have been following the same style used in the previous tests in this .feature file. But you have a point when it comes to their robustness and readability. I will update accordingly.

robons: Yeah, I want to get rid of those other ones too.

GDonRanasinghe: @robons The behave tests in cube feature now assert the cols and data instead of matching JSON representations of objects. I have updated all the tests within the cube feature file.

tests/behaviour/cube.feature
```diff
@@ -187,12 +246,47 @@ Feature: Cube!
...
+      {
+        "Dimension":"e",
+        "Value":4,
+        "Measure":"length",
+        "Unit":"cm"
+      }
+    ]
     """
 
   Scenario: Output a cube when an inline code list is defined using code list config schema v1.0, and the sort order is defined with sort object
```
robons, line 292: Whilst I appreciate that the tests are less brittle now than they were, some of the tests like these don't really test much outside of the build not failing any more. For instance, I think this test should be asserting that the sort order of the code list items has been correctly set.

tests/behaviour/cube.feature
```diff
@@ -251,10 +380,179 @@ Feature: Cube!
...
+      {
+        "Dimension":"e",
+        "Value":4,
+        "Measure":"length",
+        "Unit":"cm"
+      }
+    ]
+    """
+    
+    Scenario: Output a cube when an inline code list is defined using code list config schema v1.0 and when there are references to concepts defined elsewhere.
```
robons, line 493: This one should really be asserting that the code list CSV file which has been written to disk has the same as column containing the expected values.

tests/behaviour/cube.feature
```diff
@@ -251,10 +380,179 @@ Feature: Cube!
...
+      {
+        "Dimension":"e",
+        "Value":4,
+        "Measure":"length",
+        "Unit":"cm"
+      }
+    ]
+    """
+
+    Scenario: Output a cube when the code list is defined using code list config schema v1.0 and when there are references to concepts defined elsewhere.
```
robons, line 426: I think this one needs to be checking the code lists which have been written to file instead of the cube information itself. I wonder whether this is the right feature file for it to be part of.

tests/behaviour/cube.feature
```diff
@@ -155,12 +179,47 @@ Feature: Cube!
...
-          {'csv_column_title': 'Value', 'structural_definition': {'data_type': 'decimal', 'unit': None}, 'csv_column_uri_template': None, 'uri_safe_identifier_override': None}, 
-          {'csv_column_title': 'Measure', 'structural_definition': {'measures': [{'label': 'Length', 'description': None, 'parent_measure_uri': None, 'source_uri': None, 'uri_safe_identifier_override': None, 'arbitrary_rdf': []}]}, 'csv_column_uri_template': None, 'uri_safe_identifier_override': None}, 
-          {'csv_column_title': 'Unit', 'structural_definition': {'units': [{'label': 'cm', 'description': None, 'source_uri': None, 'uri_safe_identifier_override': None, 'arbitrary_rdf': [], 'base_unit': None, 'base_unit_scaling_factor': None, 'qudt_quantity_kind_uri': None, 'si_base_unit_conversion_multiplier': None}]}, 'csv_column_uri_template': None, 'uri_safe_identifier_override': None}
-      ]
+    [
```
robons, line 182: As I've mentioned elsewhere, I think this should really be making assertions about the contents of the code-list instead of the cube.

robons (changes requested):
I think it's moving in the right direction, the tests are definitely less brittle now, we just need to make sure we're testing the right things.

tests/behaviour/cube.feature
GDonRanasinghe: @robons The sort order is tested in the following unit tests at `test_code_list_config.py`: `test_code_list_config_model`, `test_code_list_config_model_without_schema`, `test_code_list_config_model_with_hierarchy`.

tests/behaviour/cube.feature
GDonRanasinghe: @robons The cols and their content are tested in the following unit tests at test_code_list_config.py: test_code_list_config_model, test_code_list_config_model_without_schema, test_code_list_config_model_with_hierarchy.

tests/behaviour/cube.feature
GDonRanasinghe: @robons The cols and their content are tested in the following unit tests at test_code_list_config.py: test_code_list_config_model, test_code_list_config_model_without_schema, test_code_list_config_model_with_hierarchy.

tests/behaviour/cube.feature
GDonRanasinghe: I have added a new unit test for testing the same_as field generated for the concepts defined elsewhere. See `test_code_list_config_with_concepts_defined_elsewhere` in `test_code_list_config.py`.

GDonRanasinghe:
> I think it's moving in the right direction, the tests are definitely less brittle now, we just need to make sure we're testing the right things.

@robons I have replied to your comments below outlining the unit tests as discussed.

robons (approved):

PR #544, opened 2022-07-05
Issue 528/literal attributes failing

mikeAdamss:
this: https://app.zenhub.com/workspaces/features-squad-61b72275ac896c0010a1b00b/issues/gss-cogs/csvcubed/528

After a chat with Rob have split New and Existing Attributes schemas into separate Literal and Resource variants.

Also tweaked some test names, the previously removed numbering system was differentiating some tests that had otherwise identical names, am hoping they're a bit more descriptive now.

### DIFF
--- a/src/csvcubed/readers/cubeconfig/v1/columnschema.py
+++ b/src/csvcubed/readers/cubeconfig/v1/columnschema.py
@@ -296,7 +296,7 @@ class NewAttribute(SchemaBaseClass):
                 is_required=self.required,
             )
         else:
-            if isinstance(self.values, list) or self.values:
+            if isinstance(self.values, list) and self.data_type:
                 raise Exception(
                     "Attributes cannot represent both literal values and attribute (resource) values"
                 )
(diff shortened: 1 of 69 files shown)


src/csvcubed/readers/cubeconfig/v1/columnschema.py
```diff
@@ -240,43 +240,72 @@ class AttributeValue(SchemaBaseClass):
...
 @dataclass
-class NewAttribute(SchemaBaseClass):
+class NewAttributeLiteral(SchemaBaseClass):
+    label: str
+    data_type: str
+    description: Optional[str] = None
+    from_existing: Optional[str] = None
+    definition_uri: Optional[str] = None
+    required: bool = False
+    cell_uri_template: Optional[str] = None
```
robons, line 286: Literal values (numbers, strings, etc.) don't have URIs so it should not be possible to specify a `cell_uri_template` here.

robons (commented):
Looks good for the most part, but just needs a little more attention on which models have a `cell_uri_template`.

robons (approved):

PR #549, opened 2022-07-06
adding qudt-units template+ supported it in schema

MuazzamChaud:


src/csvcubed/schema/cube-config/v1_1/schema.json
```diff
@@ -545,6 +545,13 @@
                     "enum": [
                         "observation-status"
                     ]
+                },
+                "units": {
```
robons, line 549: This isn't quite going to work. You'll need to add a `from_template` field on the `units` column which points to one of these values.

At that point, I think we should move to v1.2 since it is a significant change.

tests/behaviour/cube.feature
```diff
@@ -555,4 +554,93 @@ Feature: Cube!
         "Unit":"cm"
       }
     ]
-    """
\ No newline at end of file
+    """
+
+    Scenario: Output a cube when the sort object is not defined but the sort order is defined using schema v1.2
```
robons, line 559: why is this in you PR, isn't this work from @GDonRanasinghe ?

robons (changes requested):

robons (approved):

Issue #523, opened 2022-06-15
Upper case notations don't match up when using JSON-configured code 


robons:
When we define a code list using the JSON syntax (as in the ZIP file attached), if the notation we've defined uses capital letters, we find that the notation written to the observations.csv file has been incorrectly converted to lower case when the notation should match the upper case notations (correctly) defined in the code-list.

![image.png](https://images.zenhubusercontent.com/60d35dd0922789587fc8922d/af6bc719-de2d-4125-b9ac-4722f0a0041a
)

![image.png](https://images.zenhubusercontent.com/60d35dd0922789587fc8922d/71af81e7-bb59-439f-85bd-52aa1913c73a)

[https://app.zenhub.com/files/374709947/03797d3c-f15f-4b5f-977f-eb34e8510ce2/download](https://app.zenhub.com/files/374709947/03797d3c-f15f-4b5f-977f-eb34e8510ce2/download
)

PR #552, opened 2022-07-08
523 code list concept notation

GDonRanasinghe:
Satisfies #523.


src/csvcubed/models/cube/qb/components/concept.py
```diff
@@ -26,6 +26,9 @@ class NewQbConcept(SecondaryQbStructuralDefinition, UriIdentifiable):
 
     def get_identifier(self) -> str:
         return self.code
+    
+    def get_uri_safe_identifier(self) -> str:
```
robons, line 30: Hang on, this property is already defined on the `UriIdentifiable` type. Let's not override it.

src/csvcubed/writers/skoscodelistwriter.py
```diff
@@ -177,9 +185,15 @@ def _get_catalog_metadata(self, scheme_uri: str) -> ConceptSchemeInCatalog:
     def _get_code_list_data(self) -> pd.DataFrame:
         data_frame = pd.DataFrame(
             {
+                "Uri Identifier": [
+                    c.get_uri_safe_identifier() for c in self.new_code_list.concepts
```
robons, line 189: Just use the `uri_safe_identifier` property we already have on the Concept.

tests/unit/writers/test_skoscodelistwriter.py
```diff
@@ -46,24 +45,27 @@ def test_code_list_data_mapping():
...
         "Sort Priority",
         "Description",
     } == actual_column_names
 
     first_concept: Dict[str, Any] = data.iloc[[0]].to_dict("records")[0]
+    assert first_concept["Uri Identifier"] == "1st-20-concept"
     assert first_concept["Label"] == "First Concept"
-    assert first_concept["Notation"] == "1st-concept"
-    assert first_concept.get("Parent Notation") is None
+    assert first_concept["Notation"] == "1st-20%concept"
```
robons, line 59: Why has this changed?

GDonRanasinghe: This is because I have changed the `basic_code_list` in line 15 to include `20%` which will be detected by `uri_safe()` so we can unit test this aspect.

src/csvcubed/writers/skoscodelistwriter.py
```diff
@@ -190,11 +196,17 @@ def _get_catalog_metadata(self, scheme_uri: str) -> ConceptSchemeInCatalog:
...
         data_frame = pd.DataFrame(
             {
+                "Uri Identifier": [
+                    c.uri_safe_identifier for c in self.new_code_list.concepts
+                ],
                 "Label": [c.label for c in self.new_code_list.concepts],
                 "Notation": [c.code for c in self.new_code_list.concepts],
-                "Parent Notation": [c.parent_code for c in self.new_code_list.concepts],
+                "Parent Uri Identifier": [
+                    uri_safe(c.parent_code) if c.parent_code else None
```
robons, line 205: Instead of doing this (which I did suggest), I think we need to find the parent concept by its code and look up its `uri_safe_identifier`. This is because a user can override the parent's `uri_safe_identifier` and make it something which isn't just the `uri_safe(notation)` that it generally is.

robons (changes requested):
Apart from that new problem I've identified, it looks good.

src/csvcubed/writers/skoscodelistwriter.py
GDonRanasinghe: @robons I've updated accordingly.

robons (approved):

Issue #524, opened 2022-06-15
Add documentation on how to define code lists in the JSON format


robons:
Add a `guide` to the external documentation pages which helps a user understand how they would define both an in-line as well as external file JSON code list definition.

We need to make sure to provide a couple of examples that we can walk-through to show the user what inputs they need to provide and what the outputs will look like. 

N.B. The user should continue to use the labels as defined in their code list and these will be replaced with the notation by csvcubed. 

See the attached ZIP file for a couple of very under-developed examples of JSON code list definitions. [https://app.zenhub.com/files/374709947/9be65e1a-0030-422b-beb7-55858d7b1b71/download](https://app.zenhub.com/files/374709947/9be65e1a-0030-422b-beb7-55858d7b1b71/download)

PR #557, opened 2022-07-17
524 code list config documentation

GDonRanasinghe:
Satisfies #524.


external-docs/docs/guides/configuration/code-list-config.md
```diff
@@ -0,0 +1,195 @@
...
+
+The `sort_order` field allows sorting a code list on per concept basis.
+
+*If both the aforementioned `sort` field and the concept's `sort_order` are defined, csvcubed first sorts the concepts with `sort_order`, and then sorts the rest of the concepts according to the `sort` field. In other words, the concept's `sort_order` field has priority over the code list's `sort` field.*
+
+#### Using `same_as`
+
+The `same_as` field allows using a concept defined elsewhere in the internet (e.g. [E92 concept in Geography Linked Data](http://statistics.data.gov.uk/id/statistical-geography/E92000001)).
+
+### Example
```
robons, line 67: I've just skimmed this doc for now, but I think you're missing somewhere where you say how you point your qube-config.json's column definition at this separate file.

GDonRanasinghe: @robons I've updated the Dimension configuration table in qube-config.md file for this. Can. you check and let me know if I need to add more info to this end?

external-docs/docs/guides/configuration/code-list-config.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+
+## Defining an in-line code list
+
+This approach is recommended for defining simple code lists (e.g. code lists with a small number of concepts or simple hierarchy).
+
+As shown in the below example, an in-line code list is defined within the `qube-config.json` using the same fields discussed above. Moreover, as shown in the example, one can choose to define a `code-list-config.json` for a one dimension and define an in-line code list for another dimension.
+
+```json
```
canwaf, line 146: Is this mixed geography code list a good example here? I think a simpler example should be a scenario like "Flow" with concepts "Imports" "Exports" "Other". Covering why you would want to mix concepts could be an effective how-to scenario, since we need to explain why someone wants to do it.

canwaf (commented):
Looks excellent. I cleaned up the starting paragraphs, but think a better second example is necessary.

external-docs/docs/guides/configuration/code-list-config.md
robons: Yeah, I think I agree, if we can keep the code list as simple as possible we're probably going to make it easier for the user to understand the underlying idea we're trying to convey.

external-docs/docs/guides/configuration/code-list-config.md
robons: I think it'd be good to make it clear in this file how you use it. The qube-config.md file is a bit of a rabbit hole and I don't want user to have to go down it to understand how to define a code list. I think this document should be self-contained enough to let a user define a code list without having to refer to other documents (unless it's to understand basics).

external-docs/docs/guides/configuration/code-list-config.md
GDonRanasinghe: Thanks @canwaf @robons . I will update accordingly.

external-docs/docs/guides/configuration/code-list-config.md
```diff
@@ -0,0 +1,195 @@
...
+*If both the aforementioned `sort` field and the concept's `sort_order` are defined, csvcubed first sorts the concepts with `sort_order`, and then sorts the rest of the concepts according to the `sort` field. In other words, the concept's `sort_order` field has priority over the code list's `sort` field.*
+
+#### Using `same_as`
+
+The `same_as` field allows using a concept defined elsewhere in the internet (e.g. [E92 concept in Geography Linked Data](http://statistics.data.gov.uk/id/statistical-geography/E92000001)).
+
+### Example
+
+```json
+{
```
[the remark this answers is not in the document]
robons: Or maybe we just need something like the following above it?


```
{
  "$schema": "https://purl.org/csv-cubed/code-list-config/v1.0",
  ...metadata...,
  "sort": {
     ....sort configuration...
  },
  "concepts": [
     ....concept definitions....
  ]
}
```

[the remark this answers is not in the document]
GDonRanasinghe: > I'm wondering whether it'd be easier for the user to understand what's going on if we move the example before the explanation? What do you think?

Yes I agree - especially for the first section (i.e. Defining a code list using code-list-config.json).

Since the code list examples are now much simpler, I think we should present the examples at the beginning. I feel presenting a structure like the above is less specific/too general.

I have added the simplified examples to the beginning of each section. Please have a look at let me know what you think.

external-docs/docs/guides/configuration/code-list-config.md
```diff
@@ -0,0 +1,210 @@
...
+                        "label": "A",
+                        "description": "A data record",
+                        "notation": "a",
+                        "sort_order": 2,
+                    },
+                    {
+                        "label": "C",
+                        "description": "C data record",
+                        "notation": "c",
+                        "sort_order": 1,
```
robons, line 175: Valid JSON doesn't have trailing commas after the last property/value pair.

I'm going to change the JSON a little bit so I wouldn't worry about fixing it for now.

robons (approved):

canwaf (approved):

Issue #443, opened 2022-04-21
SkosConceptSchemeReader fails with multiple tables defined in metadata JSON-LD


robons:
Issue originally reported by @Shannon95.

When the code list information is extracted from the `industry-grouping.csv-metadata.json` file in the attached example we end up with the following exception raised: `/workspace/codelists/industry-grouping.csv-metadata.json is missing `skos:inScheme` column`.

[https://app.zenhub.com/files/374709947/e7563bb1-a921-47c6-870a-ef33c3a1d788/download](https://app.zenhub.com/files/374709947/e7563bb1-a921-47c6-870a-ef33c3a1d788/download)

The code-list definition *does* have a `skos:inScheme` column associated with it, it just so happens that the code in [skoscodelistreader.py](https://github.com/GSS-Cogs/csvcubed/blob/d5a6b2948c43c661f645198503b625795c9cac09/csvcubed/csvcubed/readers/skoscodelistreader.py#L30) only looks at the first table definition in the JSON-LD. 

- [ ] Rework [skoscodelistreader.py](https://github.com/GSS-Cogs/csvcubed/blob/d5a6b2948c43c661f645198503b625795c9cac09/csvcubed/csvcubed/readers/skoscodelistreader.py#L30)  so that it works irrespective of the number of tables defined or the order in which they are defined in the JSON-LD document.
  - It may well be useful to reuse the code used by the inspect command done by @GDonRanasinghe 
  - Make sure you write a unit test which loads the example CSV-W and ensures we can extract the relevant information from it.

PR #559, opened 2022-07-25
443 skosconceptreader fails with multi tables

GDonRanasinghe:
Satisfies #443. Includes the following works:
1. Refactored the skos code list reader to use sparql instead of json.
2. Refactored the skos code list reader to support multiple tables.
3. Refactoring sparql manager and metadataprocessor that was in inspect/ into utils.
4. Added new tests and updated failing tests.


src/csvcubed/readers/skoscodelistreader.py
```diff
@@ -27,29 +30,16 @@ def extract_code_list_concept_scheme_info(
       `concept_uri_template` uses the standard `notation` uri template variable even if the underlying file uses a
        different column name.
     """
-    table_schema_result = get_first_table_schema(code_list_csvw_path)
-    if table_schema_result is None:
-        raise ValueError(f"Unable to find tableSchema in {code_list_csvw_path}")
+    table_schema_manager = TableSchemaManager(code_list_csvw_path)
+    rdf_graph = table_schema_manager.load_json_ld_to_rdflib_graph()
```
robons, line 34: Do you think we should move this function into the `__post_init__` so the caller has less to do?

robons: Maybe the failed to load exception too?

src/csvcubed/utils/tableschema.py
```diff
@@ -1,25 +1,22 @@
 """
-Metadata Processor
+Table Schme
```
robons, line 2: `Schema`

src/csvcubed/utils/tableschema.py
```diff
@@ -144,7 +141,7 @@ def load_json_ld_to_rdflib_graph(self) -> rdflib.ConjunctiveGraph:
             raise FailedToLoadRDFGraphException(self.csvw_metadata_file_path) from ex
 
 
-def add_triples_for_file_dependencies(
+def _add_triples_for_file_dependencies(
```
robons, line 144: Can you undo this rename? The function is exposed publicly for use by csvcubed-pmd.

src/csvcubed/utils/tableschema.py
```diff
@@ -29,13 +26,13 @@
...
 )
-from csvcubed.utils.sparql import path_to_file_uri_for_rdflib
+from csvcubed.utils.sparql_handler.sparql import path_to_file_uri_for_rdflib
 
 _logger = logging.getLogger(__name__)
 
 
 @dataclass
-class MetadataProcessor:
+class TableSchemaManager:
```
robons, line 35: Think we could call this something more generic like `CsvwRdfManager` ?

tests/unit/readers/test_skoscodelistreader.py
```diff
@@ -13,70 +13,67 @@
...
+        == "http://gss-data.org.uk/data/gss_data/trade/ons-international-trade-in-services-by-subnational-areas-of-the-uk#scheme/industry-grouping"
+    )
     assert (
         concept_uri_template
-        == "http://gss-data.org.uk/def/trade/concept/age-of-business/{+notation}"
+        == "http://gss-data.org.uk/data/gss_data/trade/ons-international-trade-in-services-by-subnational-areas-of-the-uk#concept/industry-grouping/{+notation}"
     )
 
 
-def test_skos_in_scheme_missing():
```
robons, line 36: Do we want to get rid of this tests?

tests/unit/readers/test_skoscodelistreader.py
```diff
@@ -13,70 +13,67 @@
...
-    assert "is missing `url` property for code list table" in str(ex)
+    assert (
+        "Unexpected number of variables in aboutUrl Template. Expected 1, found 2"
+        in str(ex)
+    )
 
 
 def test_legacy_composite_code_list():
-    """Test that a legacy composite code list returns a sensible `aboutUrl`. Addresses bug in issue #389."""
+    """Test that a legacy composite code list returns a sensible `about-url-wriUrl`. Addresses bug in issue #389."""
```
robons, line 76: Can we also test what happens when we fail to extract the table_url/scheme url/concept url from a CSV-W? i.e. that we get an appropriate error that would make sense to a user.

tests/unit/readers/test_skoscodelistreader.py
```diff
@@ -13,70 +13,67 @@
...
+        concept_uri_template
+        == "http://gss-data.org.uk/def/trade/concept/age-of-business/{+notation}"
+    )
 
 
-def test_csv_url_missing():
+def test_num_of_variables_in_about_url_invalid_exception():
     """
-    Ensure that we get an appropriate exception when the table's URL is missing.
+    Ensures that an exception is thrown when the number of variables in about url are invalid.
```
robons, line 59: I have a vague memory of us wanting to support multiple variable names, was the code set up to allow this before?

GDonRanasinghe:
> Looking pretty good, but I have a few questions as usual.

Thanks @robons . I will get these sorted :)

src/csvcubed/utils/tableschema.py
GDonRanasinghe: Yes makes sense

tests/unit/readers/test_skoscodelistreader.py
GDonRanasinghe: @robons Actually, I tried to do this. But turned out that if either of these variables cannot be extracted, the sparql return 0 results which will trigger the `InvalidNumberOfRecordsException` exception from the python end.

Do you know whether sparql can do a validation and trigger a specific error (e.g. `aboutUrl` is missing) instead of returning 0 results?

src/csvcubed/readers/skoscodelistreader.py
GDonRanasinghe: Good shout. I have updated accordingly.

tests/unit/readers/test_skoscodelistreader.py
GDonRanasinghe: @robons I know that the `qbwriter` has some code for supporting multiple variables in the same url (see `_get_default_value_uri_for_code_list_concepts` in `qbwriter.py`). However, the existing code around `skoscodelistreader` or any code I have refactored was not supporting this as far as I can see.

tests/behaviour/steps/inspectcli.py
```diff
@@ -36,8 +36,8 @@ def step_impl(context, csv_file: str):
 
 @When("the Metadata File json-ld is loaded to a rdf graph")
 def step_impl(context):
-    metadata_processor = MetadataProcessor(context.csvw_metadata_json_path)
-    context.csvw_metadata_rdf_graph = metadata_processor.load_json_ld_to_rdflib_graph()
+    csvw_rdf_manager = CsvwRdfManager(context.csvw_metadata_json_path)
+    context.csvw_metadata_rdf_graph = csvw_rdf_manager.load_json_ld_to_rdflib_graph()
```
robons, line 40: Again this can replaced with `context.csvw_metadata_rdf_graph = csvw_rdf_manager.rdf_graph`, can't it?

tests/unit/readers/test_skoscodelistreader.py
robons: Yeah, you're right, I may just have been thinking about the `about_url` defined somewhere else.

robons (approved):
There are some small changes you can make to avoid doing things twice, but it looks good!

src/csvcubed/cli/inspect/inspect.py
```diff
@@ -33,8 +33,8 @@ def inspect(csvw_metadata_json_path: Path) -> None:
     """
     _logger.debug(f"Metadata json-ld path: {csvw_metadata_json_path.absolute()}")
 
-    metadata_processor = MetadataProcessor(csvw_metadata_json_path)
-    csvw_metadata_rdf_graph = metadata_processor.load_json_ld_to_rdflib_graph()
+    csvw_rdf_manager = CsvwRdfManager(csvw_metadata_json_path)
+    csvw_metadata_rdf_graph = csvw_rdf_manager.load_json_ld_to_rdflib_graph()
```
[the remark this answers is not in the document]
GDonRanasinghe: I have updated throughout. Thanks for pointing this out.

PR #571, opened 2022-08-16
555 code list concept uri safe identifier override, col suppressing and cell uri template features.

GDonRanasinghe:
Satisfies #555, #550 and #542.

@robons Please also create the following PURLs:
1. code list schema v1.1: https://purl.org/csv-cubed/code-list-config/v1.1
2. cube config schema v1.3: https://purl.org/csv-cubed/qube-config/v1.3

Issue #550, opened 2022-07-07
column mappings can NOT be allowed to let the user specify both cell_uri_template and values


MuazzamChaud:
change the v1_2 schema.json (@ https://github.com/GSS-Cogs/csvcubed/blob/main/src/csvcubed/schema/cube-config/v1_2/schema.json) to have the user only able to add values OR cell_uri_template to units and measures column mappings, as well as making the same change to any other columns that allow this.

This is likely a minor change since we're just fixing a bug in the existing JSON schema for a configuration that csvcubed was never able to successfully read.

- [x] Write a unit test to ensure that a schema with the new version is correctly loaded and parsed.

This bug has been triaged as low priority, however it is a relatively small alteration when done in conjunction with other Qube-Config schema changes. If you're doing any updates to the qube-config.json schema, PLEASE include this as part of your change.

GDonRanasinghe:
The new behave tests written in https://github.com/GSS-Cogs/csvcubed/pull/571 covers the tests for this change as well.

PR #572, opened 2022-08-16
550 col mappings in qube config

GDonRanasinghe:
Satisfies #550.

The new behave tests written in [#571](https://github.com/GSS-Cogs/csvcubed/pull/571) covers the tests for this change as well.

### DIFF
--- a/src/csvcubed/schema/cube-config/v1_3/schema.json
+++ b/src/csvcubed/schema/cube-config/v1_3/schema.json
@@ -68,22 +68,28 @@
                         "$ref": "#/definitions/v1.0/columnTypes/Dimension"
                     },
                     {
-                        "$ref": "#/definitions/v1.0/columnTypes/Attribute_Resource"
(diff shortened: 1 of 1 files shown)


src/csvcubed/schema/cube-config/v1_3/schema.json
```diff
@@ -189,14 +195,67 @@
...
+                        },
+                        "required": {
+                            "description": "Whether or not it is required that every observation has a value for this attribute",
+                            "type": "boolean",
+                            "default": false
                         }
                     }
                 },
-                "Dimension": {
+                "Dimension_New": {
```
robons, line 258: This breaks things.

robons (approved):
I've made some changes which correct some small mistakes and making it clear that you should only be able to set the `qudt` units template in an existing units column. 

Please make sure you review and test your changes when they get to PR stage.

PR #576, opened 2022-08-20
Canwaf/issue570

mikeAdamss:
canwaf's  fix for the out/out bug https://app.zenhub.com/workspaces/features-squad-61b72275ac896c0010a1b00b/issues/gss-cogs/csvcubed/570

also includes some changes around buffering sub processes that were causing issues for the BDD test on this PR.

there's only so much you can pass to a subprocess `PIPE`, they're very resource constrained. Trying to pipe the large JsonSchema validation failure command line output to stdout was resulting in deadlocks on my personal machine, I _believe_ that may been the cause of the eroneous failed imports earlier on this branch (import a library requires resource) but it's hard to be 100% on that, regardless everything works now because ...

I've updated the subprocess handling to use a temp file as the stdout as stderr then read them in after the process is complete, that effectively gives us an inexhaustible buffer, seems to have done the trick and proofs us against these kinds of issues in the future.


tests/behaviour/steps/cli.py
```diff
@@ -87,15 +87,30 @@ def step_impl(context):
...
+
+    # Use temp files not a PIPE, a PIPE has a tiny buffer than
+    # can deadlock or result in eroneous resource exhaustion behaviour 
+    # where encountering some of our larger outputs (jsonSchemaErrors result
+    # in large writes to stdout)
+    Path(tmp_dir_path / "buffer").mkdir()
+    stdout_path = Path(tmp_dir_path / "buffer" / "stdout")
+    stderr_path = Path(tmp_dir_path / "buffer" / "stderr")
+
+    with open(stdout_path, "w") as stdout_file:
```
robons, line 99: You can do multiple `with` bindings on the same line (so you don't have to indent twice) 

e.g. 

```python
with A() as a, B() as b:
    suite
```

https://docs.python.org/3.1/reference/compound_stmts.html#with

tests/behaviour/steps/cubes.py
```diff
@@ -119,7 +119,7 @@ def step_impl(context):
             data_file,
             config_file,
             output_directory=context.out_dir,
-            validation_errors_file_out=context.out_dir / "validation_errors.json",
+            validation_errors_file_name=context.out_dir / "validation_errors.json",
```
robons, line 122: Shouldn't this just be the name of the file rather than including the out directory too?

tests/test-cases/configloaders/qube_validation_error.json
```diff
@@ -0,0 +1,25 @@
+{
+    "$schema": "https://purl.org/csv-cubed/qube-config/v1",
+    "id": "ons-international-trade-in-services-by-subnational-areas-of-the-uk",
+    "title": "International trade in services by subnational areas of the UK",
+    "description": "Experimental estimated value of exports and imports of services for 2018 for NUTS1, NUTS2, NUTS3 and 15 city regions, including industry and non-EU and EU split.",
+    "landingPage": "https://www.ons.gov.uk/businessindustryandtrade/internationaltrade/datasets/internationaltradeinservicesbysubnationalareasoftheuk",
+    "publisher": "https://example.com",
+    "published": "2020-09-10",
+    "families": [
```
robons, line 9: This configuration is from an info.json style file. Is there a reason this is in a qube-config.json file?

tests/unit/readers/cubeconfig/v1_0/test_cube_config_extra_cols.py
```diff
@@ -23,7 +23,7 @@ def test_build_with_fail():
                 output_directory=output,
                 csv_path=csv,
                 fail_when_validation_error_occurs=True,
-                validation_errors_file_out=Path(output, "validation_errors.json"),
+                validation_errors_file_name=Path(output, "validation_errors.json"),
```
robons, line 26: Again, shouldn't this just be the file name now? There are other lines like this in the same file.

tests/unit/readers/cubeconfig/v1_0/test_cube_from_config_json.py
```diff
@@ -69,7 +69,7 @@ def test_build():
             output_directory=output,
             csv_path=csv,
             fail_when_validation_error_occurs=True,
-            validation_errors_file_out=Path(output, "validation_errors.json"),
+            validation_errors_file_name=Path(output, "validation_errors.json"),
```
robons, line 72: Again, shouldn't this just be the file name now? There are other lines like this in the same file.

tests/unit/readers/cubeconfig/v1_0/test_cube_from_convention.py
```diff
@@ -49,7 +49,7 @@ def test_01_build_convention_ok():
             output_directory=output,
             csv_path=csv,
             fail_when_validation_error_occurs=False,
-            validation_errors_file_out=Path(output, "validation_errors.json"),
+            validation_errors_file_name=Path(output, "validation_errors.json"),
```
robons, line 52: Again, shouldn't this just be the file name now?

robons (approved):
Mostly good, just check those tests are passing in an appropriate argument for the validation errors file name.

tests/behaviour/steps/cli.py
mikeAdamss: I had no idea that was a thing, cool :smile:

tests/test-cases/configloaders/qube_validation_error.json
mikeAdamss: gonna tag in @canwaf , I only really got behave working

mikeAdamss: think I follow this, we're testing the JsonSchema validation here, if I remove $schema or unit, or measure or any of the additional things then we fail the "cli will succeed" step and don't get to the schema validation that we're after.

robons: JSON Schema validation never causes the CLI to fail. It just provides warnings to the user (which we call schema validation errors)

mikeAdamss: yeah I got that, turned out it was all because we had a step `Then the csvcubed CLI should succeed` which it didn't without the extra info (we'd get a failed status code from say a non schema based "must have a unit" validation check).

removing that step let me strip down the input, I think its more like what you had in mind now.

Issue #445, opened 2022-04-21
If unit is defined based on an existing unit and no scaling factor set, ensure that scaling factor is set to one.


robons:
In PR #442 I removed the vaildation ensuring that if a user specified an existing unit which theirs should be based on, they were required to specify the scaling factor between their unit and the existing one. This behaviour was designed to ensure we created information which was as clear as possible relating new units to the ones they are based on, however it isn't as convenient as it should be for users.

- [x] Once complete, restore the test removed from PR#442.

If a user specifies an existing unit to reuse, they automatically extend it (without knowing) so that they can set new labels, descriptions, etc. If the user wishes to, they should be able to specify the scaling factor (if relevant), but if they don't provide it we should just assume that it is `1` (i.e. there is a one-to-one relationship between the existing unit and their new more specialised unit). This behaviour can be implemented in a pydantic validator (they are allowed to change values as well as just throw exceptions!)

- Ensure you write unit/behaviour tests to:
  - [x] Ensure a user can extend an existing unit without specifying the scaling factor - this should be set to 1
  - [x] Ensure a user can extend an existing unit and specify a scaling factor (make sure it's not 1 when you test this) and that this information is correctly retained.

## Design consideration - interface specific

This change should only impact the reading of the `qube-config.json`, not the underlying implementation of the objects `NewQbUnit` or `ExistingQbUnit`.

GDonRanasinghe:
@robons Can you please provide me with an input example for this?

robons:
> @robons Can you please provide me with an input example for this?

So say you have the following observations column configuration:

```json
        "Population": {
            "type": "observations",
            "unit": {
                "label": "People",
                "from_existing": "http://qudt.org/vocab/unit/NUM"
            }
        }
```

This essentially means we're defining a new unit called `People` based of the `qudt:NUM` (Number) unit. But we've neglected to set `"scaling_factor": 1` (which it really is), so we want the application to act as if you've set the scaling factor to 1 given that you've provided an existing unit this new one is based on.

PR #579, opened 2022-08-24
Scaling factor of exiting units

GDonRanasinghe:
Satisfies #445 .


src/csvcubed/readers/cubeconfig/v1/columnschema.py
```diff
@@ -496,6 +498,18 @@ def _map_attribute_values(
     return new_attribute_values
 
 
+def _get_unit_scaling_factor(unit: Unit) -> Optional[float]:
+    """
+    If the user wishes to, they should be able to specify the scaling factor (if relevant), but if they don't provide it we should just assume that it is 1 (i.e. there is a one-to-one relationship between the existing unit and their new more specialised unit). If the unit is not derived from an existing unit, there will be not scaling factor.
```
robons, line 503: This is a looooong line. Didn't your linter break it for you?

tests/unit/cube/qb/test_qb_unit.py
```diff
@@ -42,6 +50,91 @@ def test_new_unit_base_unit_validation():
...
+    with TemporaryDirectory() as t:
+        temp_dir = Path(t)
+
+        cube_config = {
+            "columns": {
+                "Amount": {
+                    "type": "observations",
+                    "unit": {
+                        "label": "Count",
+                        "from_existing": "http://qudt.org/vocab/unit/NUM",
```
robons, line 106: We could probably use a third test to ensure that if you specify a new unit without basing it off an existing one then you end up with no scaling factor defined.

At that point you may want to refactor the unit test into a function like the following:

```python
unit = _test_unit_definition({ 
  "label": "Count",
  "from_existing": "..."
})
assert isinstance(unit, NewQbUnit)
assert unit.base_unit_scaling_factor ....
```

edit: I've just seen the following file which pretty much covers that base - there's a fair bit of overlap between the two set of tests.

robons (approved):

Issue #564, opened 2022-07-28
inspect command not working properly on new style CSV-W code lists


GDonRanasinghe:
It looks like the inspect command is erroring when handling parent-child hierarchies at the metadataprinter.py.

- [x] apply a suitable solution to fix the issue.
- [x] update any related unit and behave tests.
- [x] We need to be able to `inspect` both the older style code-lists using `notation` as the identifier as well as the newer style code lists wiith a separate uri identifier field. We need to test both of these.

Try and find a general solution to this that doesn't rely on checking whether a column with a specific name exists in the CSV-W -- we want to make sure we're not writing brittle code that will break whenever we change the CSV-W's format again.

The error is as below and the input examples are attached [https://app.zenhub.com/files/374709947/77ba7039-e027-453a-aec3-f083161625a8/download](https://app.zenhub.com/files/374709947/77ba7039-e027-453a-aec3-f083161625a8/download).

```
rob_barry@cloudshell:~$ csvcubed version
v0.1.6
rob_barry@cloudshell:~$ csvcubed inspect out/mixed-geographies-code-list.csv-metadata.json
2022-07-21 17:14:53,630 - csvcubed.cli.entrypoint - CRITICAL - Traceback (most recent call last):
  File "/home/rob_barry/.local/lib/python3.9/site-packages/csvcubed/cli/entrypoint.py", line 123, in inspect_command
    inspect(csvw_metadata_json_path)
  File "/home/rob_barry/.local/lib/python3.9/site-packages/csvcubed/cli/inspect/inspect.py", line 60, in inspect
    ) = _generate_printables(
  File "/home/rob_barry/.local/lib/python3.9/site-packages/csvcubed/cli/inspect/inspect.py", line 93, in _generate_printables
    metadata_printer = MetadataPrinter(
  File "<string>", line 6, in __init__
  File "/home/rob_barry/.local/lib/python3.9/site-packages/csvcubed/cli/inspect/metadataprinter.py", line 213, in __post_init__
    self.generate_codelist_results()
  File "/home/rob_barry/.local/lib/python3.9/site-packages/csvcubed/cli/inspect/metadataprinter.py", line 203, in generate_codelist_results
    self.result_concepts_hierachy_info = get_concepts_hierarchy_info(
  File "/home/rob_barry/.local/lib/python3.9/site-packages/csvcubed/cli/inspect/inspectdatasetmanager.py", line 214, in get_concepts_hierarchy_info
[...]

PR #580, opened 2022-08-30
inspect command update to match new code lists

GDonRanasinghe:
Satisfies #564 . The update involves using the `primaryKey` in the table json (retrieved using sparql) to identify the notation column in the code list config.


canwaf (approved):
Offt.

src/csvcubed/utils/sparql_handler/sparql_queries/select_codelist_primary_key_by_dataset_url.sparql
```diff
@@ -0,0 +1,18 @@
...
+PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+
+
+SELECT DISTINCT ?tablePrimaryKey
+WHERE {   
+
+    ?table csvw:tableSchema ?tableSchema;
+           csvw:url ?tableUrl.
+           
+    ?tableSchema csvw:primaryKey ?tablePrimaryKey.
```
robons, line 13: What happens if the primary key is composed of more than one column?

src/csvcubed/models/csvcubedexception.py
```diff
@@ -94,6 +98,10 @@ class CsvcubedExceptionUrls(Enum):
         "http://purl.org/csv-cubed/err/column-definition-not-supported"
     )
 
+    PrimaryKeyColumnTitleCannotBeNone = (
+        "http://purl.org/csv-cubed/err/invalid_title_for_code_list_col_with_primary_key"
```
robons, line 102: Can you make this shorter and use dashes to be consistent with the others?

Perhaps `http://purl.org/csv-cubed/err/invalid-pk-col-title`?

src/csvcubed/cli/inspect/metadataprinter.py
```diff
@@ -111,19 +115,20 @@ def get_dataset_url(
...
+        parent_notation_col_title = get_codelist_col_title_by_property_url(
             columns, CodelistPropertyUrl.SkosBroader
         )
-        label_col_name = get_codelist_col_title_by_property_url(
+        label_col_title = get_codelist_col_title_by_property_url(
             columns, CodelistPropertyUrl.RDFLabel
         )
-        notation_col_name = get_codelist_col_title_by_property_url(
-            columns, CodelistPropertyUrl.SkosNotation
+        notation_col_title = get_codelist_notation_col_title_from_primary_key(
```
robons, line 127: I wouldn't call it `notation` anymore. Perhaps `unique_identifier`?

src/csvcubed/models/sparqlresults.py
```diff
@@ -219,6 +220,15 @@ class CodeListColsByDatasetUrlResult:
     columns: List[CodelistColumnResult]
 
 
+@dataclass
+class CodeListPrimaryKeyByDatasetUrlResult:
+    """
+    Model to represent select codelist primary key by table url.
+    """
+
+    primary_key: str
```
robons, line 229: Again, what happens if a primary key is a composite of multiple columns?

src/csvcubed/cli/inspect/metadataprinter.py
```diff
@@ -195,13 +200,22 @@ def generate_codelist_results(self):
         self.result_code_list_cols = select_codelist_cols_by_dataset_url(
             self.csvw_metadata_rdf_graph, self.dataset_url
         )
+        # Retrieving the primary key of the code list to identify the notation column
```
robons, line 203: Again, we can't call it the `notation` column anymore since there's another column in the CSV-W which holds the notation.

src/csvcubed/utils/skos/codelist.py
```diff
@@ -63,6 +65,32 @@ def get_codelist_col_title_by_property_url(
     return results[0].column_title
 
 
+def get_codelist_notation_col_title_from_primary_key(
```
robons, line 68: Again, we can't call this the `notation` column anymore. It's really the unique identifier column.

src/csvcubed/utils/sparql_handler/sparql_queries/select_codelist_cols_by_dataset_url.sparql
```diff
@@ -4,20 +4,25 @@ PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
...
 
-SELECT DISTINCT ?columnPropertyUrl ?columnValueUrl ?columnTitle
+SELECT DISTINCT ?columnPropertyUrl ?columnValueUrl ?columnTitle ?columnName
 WHERE {   
 
     ?table csvw:url ?tableUrl.
     ?table csvw:tableSchema/csvw:column/rdf:rest*/rdf:first ?column.
 
-    ?column csvw:propertyUrl ?columnPropertyUrl.
+    OPTIONAL{
```
robons, line 13: So this change is in here because we want to bring out the unique identifier column which is suppressed and doesn't have a property URL?

src/csvcubed/utils/sparql_handler/sparqlmanager.py
```diff
@@ -368,6 +374,33 @@ def select_codelist_cols_by_dataset_url(
...
+        init_bindings={"table_url": Literal(table_url)},
+    )
+
+    if len(results) != 1:
+        raise InvalidNumberOfRecordsException(
+            record_description=f"result for the {SPARQLQueryName.SELECT_CODELIST_PRIMARY_KEY_BY_DATASET_URL.value} sparql query",
+            excepted_num_of_records=1,
+            num_of_records=len(results),
+        )
+    return map_codelist_primary_key_by_dataset_url_result(results[0])
```
robons, line 401: This needs more thought about what happens when the primary key is a composite of multiple columns.

tests/unit/utils/skos/test_codelist.py
```diff
@@ -58,8 +60,12 @@ def test_get_codelist_col_title_by_property_url_notation():
     result_code_list_cols = select_codelist_cols_by_dataset_url(
         csvw_metadata_rdf_graph, dataset_url
     )
-    notation_col_name = get_codelist_col_title_by_property_url(
-        result_code_list_cols.columns, CodelistPropertyUrl.SkosNotation
+    result_code_list_primary_key = select_codelist_primarykey_by_dataset_url(
```
robons, line 63: Do we have any new tests which explicitly test that this works with the new csvcubed outputs?

robons (commented):
We need to rename a bunch of locations where we're referring to the `notation` column. There's still a notation column in these CSVs but we're looking for the uniquely identifying column now!

We also need to have some explicit thought about what (if anything) we can do where a code-list might have multiple primary keys. I know our outputs don't currently have them, but I'd prefer having generic functionality which could pull out the columns associated with a primary key (so we can use it elsewhere). 

We could also use a unit/integration test which explicitly tests inspecting one of the new style inputs.

src/csvcubed/utils/sparql_handler/sparql_queries/select_codelist_cols_by_dataset_url.sparql
GDonRanasinghe: @robons Correct

src/csvcubed/utils/sparql_handler/sparql_queries/select_codelist_primary_key_by_dataset_url.sparql
GDonRanasinghe: As discussed, composite primary keys are now handled.

tests/unit/utils/skos/test_codelist.py
GDonRanasinghe: The new tests is added to behave tests.

src/csvcubed/cli/inspect/metadataprinter.py
```diff
@@ -195,13 +200,29 @@ def generate_codelist_results(self):
         self.result_code_list_cols = select_codelist_cols_by_dataset_url(
             self.csvw_metadata_rdf_graph, self.dataset_url
         )
+        # Retrieving the primary key of the code list to identify the unique identifier
+        result_primary_keys_by_dataset_url: PrimaryKeysByDatasetUrlResult = (
```
robons, line 204: I'd call it primary key columns instead of keys, keys implies there are multiple different indexes/unique values.

src/csvcubed/cli/inspect/metadataprinter.py
```diff
@@ -195,13 +200,29 @@ def generate_codelist_results(self):
...
+            )        
         (
-            parent_notation_col_name,
-            label_col_name,
-            notation_col_name,
-        ) = self.get_parent_label_notation_col_names(self.result_code_list_cols.columns)
+            parent_notation_col_title,
+            label_col_title,
+            notation_col_title,
+        ) = self.get_parent_label_notation_col_titles(
```
robons, line 221: Don't forget, this shouldn't be called `notation`/`parent_notation` anymore since the columns don't necessarily represent the notation at all (but there is still notation information in the CSV-W)

src/csvcubed/utils/skos/codelist.py
```diff
@@ -63,6 +65,32 @@ def get_codelist_col_title_by_property_url(
     return results[0].column_title
 
 
+def get_codelist_unique_identifier_from_primary_key(
```
robons, line 68: Can you be clear it's actually the unique identifier's column title you're getting here? Or perhaps make this function more general so that it's clear that it's actually just taking a column's name and returning the column title?

src/csvcubed/utils/skos/codelist.py
```diff
@@ -80,18 +108,15 @@ def build_concepts_hierarchy_tree(
     tree = Tree()
     tree.create_node("root", identifier="root")
 
-    # Replacing empty values with None and sorting consepts by Sort Priortiy to maintain the order when iterating below.
+    # Replacing empty values with None.
```
robons, line 111: Do we not need to apply the sort order anymore?

src/csvcubed/utils/sparql_handler/sparqlmanager.py
```diff
@@ -367,6 +373,26 @@ def select_codelist_cols_by_dataset_url(
         )
     return map_codelist_cols_by_dataset_url_result(results)
 
+def select_primary_keys_by_dataset_url(
```
robons, line 376: Again, can you make it clear that you're selecting the primary key column names?

tests/behaviour/inspectcli.feature
```diff
@@ -750,4 +750,61 @@ Feature: Behaviour testing of csvcubed inspect.
 
     Scenario: inspect should output error when the metadata json-ld input does not exist
         Given a none existing test-case file "cli/inspect/not_exists.csv-metadata.json"
-        Then the file not found error is displayed "Could not find test-case file"
\ No newline at end of file
+        Then the file not found error is displayed "Could not find test-case file"
+
+    Scenario: inspect should produce the expected printable for code list metadata json-ld with concepts that uses uri safe identifier override.
```
robons, line 755: The purpose of the test is to check that the newer-style code-list outputs from csvcubed can be inspected properly. It isn't specifically the URI safe identifer override changes we need to test here.

src/csvcubed/utils/skos/codelist.py
GDonRanasinghe: @robons I had a look at this. Sorting before constructing the tree wasn't actually needed in the first place. I believe I did this before the sort-order and sort-priority bug fixes and left it there. So this was always a redundant functionality.

tests/behaviour/inspectcli.feature
GDonRanasinghe: The two tests are now added.

src/csvcubed/utils/skos/codelist.py
```diff
@@ -63,6 +65,32 @@ def get_codelist_col_title_by_property_url(
     return results[0].column_title
 
 
+def get_codelist_col_title_from_col_name(
```
robons, line 68: Does this genuinely only work for code lists?

src/csvcubed/utils/sparql_handler/sparqlmanager.py
```diff
@@ -367,6 +373,26 @@ def select_codelist_cols_by_dataset_url(
         )
     return map_codelist_cols_by_dataset_url_result(results)
 
+def select_primary_key_col_names_by_dataset_url(
```
robons, line 376: Can you replace `dataset_url` with `csv_url` here? I believe that's what we're matching on.

tests/test-cases/cli/inspect/uri_safe_identifier_override/code-list-with-uri-safe-identifier-override.csv
```diff
@@ -0,0 +1,6 @@
+Uri Identifier,Label,Notation,Parent Uri Identifier,Sort Priority,Description
```
robons, line 1: Are we using any of these files in the `uri_safe_identifier_override` folder anymore? If not, can you remove them please?

robons (approved):
If you could just address the comments I've left before merging.

PR #594, opened 2022-09-16
Sonarcloud GitHub action

MuazzamChaud:


robons (changes requested):
I'm not sure why, but you've ended up with a load of random other changes in this PR. Make sure you remove them before we merge this.

PR #616, opened 2022-10-05
518 codelist same as bug

GDonRanasinghe:
Fixed the output csv not having the same as uri column when the concepts defined elsewhere are used.


tests/unit/models/test_code_list_config.py
```diff
@@ -166,3 +182,68 @@ def test_code_list_config_with_concepts_defined_elsewhere():
...
+        output_dir = temp_dir / "out"
+        config_path = _test_case_writers_base_dir / "skoscodelistwriter" / "config.json"
+        csv_path = _test_case_writers_base_dir / "skoscodelistwriter" / "data.csv"
+
+        cli_build(
+            output_directory=output_dir, csv_path=csv_path, config_path=config_path
+        )
+
+        output_df = pd.read_csv(output_dir / "geography.csv")
+        assert "Original Concept URI" in output_df.columns
```
robons, line 249: Can you make sure the URIs you're expecting get serialised?

robons (approved):
Approved on the condition that you test that the same as URIs defined in the JSON file are definitely in the data frame/CSV.

canwaf (changes requested):
This change has regressed the functionality for the `from_template: year` configuration option.

Additionally, you have put live data in your tests. Please synthesise more narrow test cases for unit tests.

I need to have the from_template functionality working. Let's talk about how the serialisation has changed and how we can improve our test coverage to prevent future regression from getting to this point.

canwaf (approved):
Thank you @CharlesRendle for making generic tests for this; and only testing one instance (since we only need to test things once).

Issue #614, opened 2022-10-03
prov-o csvcubed version number is harded coded in behaviour tests


CharlesRendle:
When creating the release tag v0.1.10-rc1 the behaviour tests failed (skoscodlstwriter + qbwriter) for the following reason:

The behaviour tests for ticket (#541), we used the .dev0 tag which will fail when run with any other tag number.

We need to re-work this test so that any tag is recognised by the output rdf.

- [x] Add a step to the end of each of those behaviour tests that makes sure the following triples are present.
   - The URI should be generated in the step definition using the helper function already defined in `csvcubed.utils.version`.

```
basicCodeList:csvcubed-build-activity a prov:Activity;
      prov:used <https://github.com/GSS-Cogs/csvcubed/releases/tag/v0.1.0.dev0>.
```

PR #617, opened 2022-10-07
Issue 614 - prov-o csvcubed version number is harded coded in behaviour tests

CharlesRendle:
Adresses ticket #614

Previously, the two behaviour tests which verified that the csvcubed version specific metadata triples has been serialised correctly to the resultant RDF contained a hardcoded value for the version URI which caused the tests to fail when running under any other tag.

This PR has changed these two tests by adding an additional step to each of their respective scenarios which construct the correct triples using the version util function created in the [original ticket](https://github.com/GSS-Cogs/csvcubed/issues/541).

The assertion uses the same devtools function as the step `and the RDF should contain` but this couldn’t be wholly reused for this fix because we needed to add the version function call to the step definition instead of calling from `context.text`.

The new step definition for `and the RDF should contain version specific triples` has been placed in `qbwriter` but I can understand if it would be more appropriate to have built this step definition in `csvcabed-devtools`.

### DIFF
--- a/tests/behaviour/steps/qbwriter.py
+++ b/tests/behaviour/steps/qbwriter.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from behave import Given, When, Then, Step
 from csvcubeddevtools.behaviour.file import get_context_temp_dir_path
 from csvcubeddevtools.helpers.file import get_test_cases_dir
+from csvcubeddevtools.behaviour.rdf import test_graph_diff
 from rdflib import Graph
 
 from csvcubed.models.cube import *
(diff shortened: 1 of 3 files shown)


tests/behaviour/steps/qbwriter.py
```diff
@@ -891,3 +893,13 @@ def assert_uri_style_for_uri(uri_style: URIStyle, uri: str, node):
         assert path.endswith(".csv") or path.endswith(
             ".json"
         ), f"expected {node} to end with .csv or .json"
+
+@then(u'the RDF should contain version specific triples')
+def step_impl(context):
+
+    context.version_triples = context.text + "\n" + f"prov:used <{get_csvcubed_version_uri()}> ."
```
GDonRanasinghe, line 900: We should use `os.linesep` instead of `\n` so that new lines will be recognised in Windows.

GDonRanasinghe (changes requested):
@CharlesRendle Just a one change request on the new line.

canwaf (approved):

GDonRanasinghe (approved):

PR #619, opened 2022-10-11
Fixing a bug in windows testing (regarding QbWriter)

nimshi89:
rdflib raised a syntax error due to the "\r" char that was being picked up on the windows machine.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ packages = [
   {include = "csvcubed", from = "src"},
 ]
 readme = "README.md"
-version = "0.1.11"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -8,7 +8,7 @@ packages = [
   {include = "csvcubed", from = "src"},
 ]
 readme = "README.md"
-version = "0.1.0-dev"
+version = "0.1.11"
```
robons, line 11: This needs to be removed from the change before merging in to main.

robons (approved):
Yay

Issue #607, opened 2022-09-27
Inspect Command - Add column to link attribute, measures and units cols to the observations column they describe


robons:
* [x] Add an extra column to the *columns table* to display which observations column any attribute, measures and units columns are linked with; this will require modifications to the SPARQL query. 
  ![image.png](https://images.zenhubusercontent.com/60d35dd0922789587fc8922d/48750015-12ec-453b-a560-5fcf1966248f) 
  * This needs to be designed to work with (and be tested with) both a multi-measure pivoted CSV-W input **and** a single-measure pivoted CSV-W input.
* [x] Add a new unit (a behave test matching on text is not desirable here) test with a single-measure pivoted CSV-W input generated by csvcubed. Ensure that the *columns table* correctly links units, measure dimension, and attribute columns to the single obs val column. 
* [x] Add a new unit test (a behave test matching on text is not desirable here) with a pivoted multi-measure output from csvcubed. 
  - Ensure that:
    * [x] at least one of the observations columns defines the unit directly against the observation values column. 
    * [x] at least one of the observations columns has a linked units column (i.e. the unit is not defined directly against the observations column). Ensure that the units vary a bit within this column.
    * [x] there are at least 20 rows in the CSV file.
    * [x] at least one of the observations columns has a linked attribute column.
  - Test that:
     * [x] ensure that the *columns table* correctly links units + attributes columns to their observations column.  
     * [x] we don't see `First 10 Observations`/`Last 10 Observations` and instead see whatever text you made it output.
* [x] Alter the bits which say `First 10 Observations` and `Last 10 Observations` to say something appropriate when a pivoted mult-measure input is present. Perhaps `First 10 Rows`? Also make sure that whatever it displays when there are fewer than 10 rows is sensible with the new inputs too.
 * [x] Add a new unit (a behave test matching on text is not desirable here) test with a pivoted multi-measure output from csvcubed with <10 rows
   - Test that the equivalent of the `First 10 Observations` functionality says the sensible thing you configured it to say.

**Ensure that you don't alter any existing unit/behaviour tests since this is additional functionality.**

N.B. We want tomove from behaviour testing text strings to be able to unit test the python models (which will be part of our future API) and decoupling the testing of string formatting/text output from the testing that our queries have correctly interpreted by the API. 

[...]

PR #627, opened 2022-10-26
607 inspect command columns update

GDonRanasinghe:
Satisfies #607 .


src/csvcubed/models/sparqlresults.py
```diff
@@ -582,3 +663,21 @@ def map_table_schema_properties_result(
...
+def map_is_pivoted_shape_for_measures_in_data_set(
+    sparql_results: List[ResultRow],
+) -> List[IsPivotedShapeMeasureResult]:
+    """
+    Maps the sparql query result to objects of type IsPivotedMeasureResult that are then returned.
+    """
+    is_pivoted_shape_measure_results = list(
+        map(
+            lambda result: IsPivotedShapeMeasureResult(
+                measure=str(result.asdict()["measure"]),
```
robons, line 677: Again, can you make sure to only call `.asdict` once per result given that I think this is likely an expensive operation?

src/csvcubed/utils/qb/cube.py
```diff
@@ -17,6 +17,7 @@
     QbObservationValue,
     QbColumnStructuralDefinition,
 )
+from csvcubed.utils.sparql_handler.sparqlmanager import CSVWShape
```
robons, line 20: Can you store this elsewhere? I'd like to make sure that the `csvcubed build` functionality isn't dependent on `csvcubed inspect` functionality and vice versa. If you place it somewhere which isn't in one or the other we can avoid having any dependencies there.

src/csvcubed/utils/qb/cube.py
```diff
@@ -91,3 +92,29 @@ def get_all_units(cube: Cube) -> Set[QbUnit]:
...
+            all_standard_shape
+            and not obs_val_col.structural_definition.is_pivoted_shape_observation
+        )
+
+    if all_pivoted:
+        return CSVWShape.Pivoted
+    elif all_standard_shape:
+        return CSVWShape.Standard
+    else:
+        raise TypeError("The cube cannot be in both standard and pivoted shape")
```
robons, line 120: Could you make it clear in this error message that it's about some obs val columns being in pivoted shape and others not?

src/csvcubed/utils/sparql_handler/sparqlmanager.py
```diff
@@ -51,6 +53,18 @@
 _logger = logging.getLogger(__name__)
 
 
+class CSVWShape(Enum):
```
robons, line 56: Can we call this `CubeShape` for consistency? A code list will never be in either of these shapes.

tests/behaviour/steps/inspectcli.py
```diff
@@ -1,12 +1,37 @@
+import pandas as pd
 from difflib import ndiff
 from pathlib import Path
 
 from behave import *
+from pandas.util.testing import assert_frame_equal
+from more_itertools import first
```
robons, line 7: We already have a function called `first` defined [here](https://github.com/GSS-Cogs/csvcubed/blob/main/src/csvcubed/utils/iterables.py). Can you use that instead so we don't add any new dependencies, even if they are only in testing.

tests/behaviour/steps/inspectcli.py
```diff
@@ -59,9 +90,11 @@ def step_impl(context):
 def step_impl(context):
     metadata_printer = MetadataPrinter(
         context.csvw_type,
+        context.csvw_shape,
         context.csvw_metadata_rdf_graph,
         context.csvw_metadata_json_path,
     )
+    # TODO: Remove below once all the tests are updated to not match strings
```
robons, line 97: Do we need this still?

tests/behaviour/steps/inspectcli.py
```diff
@@ -81,12 +114,26 @@ def step_impl(context):
         and context.dataset_observations_info_printable
         and context.dataset_val_counts_by_measure_unit_info_printable
     )
+    # TODO: Remove above once all the tests are updated to not match strings
```
robons, line 117: And again?

tests/behaviour/steps/inspectcli.py
```diff
@@ -140,6 +187,7 @@ def step_impl(context):
 
 @Then("the Dataset Information Printable should be")
 def step_impl(context):
+    print(context.dataset_observations_info_printable)
```
robons, line 190: Can you remove this?

tests/behaviour/steps/inspectcli.py
```diff
@@ -171,3 +219,341 @@ def step_impl(context, error_message: str):
...
+    assert (
+        len(result_catalog_metadata.themes) == 1
+        and result_catalog_metadata.themes[0] == ""
+    )
+    assert (
+        len(result_catalog_metadata.keywords) == 1
+        and result_catalog_metadata.keywords[0] == ""
+    )
+    assert result_catalog_metadata.contact_point == "None"
+    assert result_catalog_metadata.identifier == "qb-id-10004"
```
robons, line 255: Should this be an input to the step definition? 

Can you either reword the step definition so it takes all the values which might change as parameters, or make the step name so specific no one else will attempt to reuse it?

tests/behaviour/steps/inspectcli.py
```diff
@@ -171,3 +219,341 @@ def step_impl(context, error_message: str):
...
+    result_qube_components: QubeComponentsResult = context.result_qube_components
+    assert result_qube_components is not None
+
+    components = result_qube_components.qube_components
+    assert len(components) == 5
+
+    component = get_dsd_component_by_property_url(
+        components, "qb-id-10004.csv#dimension/some-dimension"
+    )
+    assert_dsd_component_equal(
```
robons, line 273: These tests are so specific I'm not sure they really fit into the behaviour testing framework. 

Could you convert them into unit tests? Unless we absolutely need any of the docker-based functionality.

tests/unit/utils/sparqlhandler/test_sparqlmanager.py
```diff
@@ -159,32 +241,203 @@ def test_select_csvw_catalog_metadata_for_codelist():
     assert result.identifier == "Alcohol Content"
 
 
-def test_select_csvw_dsd_dataset():
+def test_select_csvw_dsd_dataset_for_pivoted_multi_measure_data_set():
```
robons, line 244: Do we need to test this for the old-style single measure pivoted input? The inspect functionality should still be capable of querying and interpreting it, even if the corresponding version of csvcubed can't output it anymore.

src/csvcubed/cli/inspect/metadatainputvalidator.py
```diff
@@ -41,20 +43,17 @@ class MetadataValidator:
...
+        Detects the validity, type and shape of the csvw.
         """
-        metadata_type = self._detect_type()
-        return (
-            metadata_type == CSVWType.QbDataSet or metadata_type == CSVWType.CodeList,
-            metadata_type,
-        )
+        csvw_type = self._detect_type()
+        csvw_shape = self._detect_shape() if csvw_type == CSVWType.QbDataSet else None
+        validity = csvw_type == CSVWType.QbDataSet or csvw_type == CSVWType.CodeList
```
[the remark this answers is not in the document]
GDonRanasinghe: You are correct. We have updated it to throw an exception.

src/csvcubed/cli/inspect/metadatainputvalidator.py
```diff
@@ -78,3 +77,22 @@ def _detect_type(self) -> CSVWType:
...
         else:
             return CSVWType.Other
+
+    def _detect_shape(self) -> CSVWShape:
+        """
+        Given a metadata validator as input, returns the shape of the cube that metadata describes (Pivoted or Standard).
+        """
+        all_pivoted = True
+        all_standard_shape = True
+        for measure in self.is_pivoted_measures:
```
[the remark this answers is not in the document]
GDonRanasinghe: @robons We have updated this as follows:
1. The `self.is_pivoted_measures` is not a class property now.
2. The sparql for detecting pivoted measures is run in the inspect.py class. This is similar to how the other sparqls are being run from this class.

tests/behaviour/steps/inspectcli.py
GDonRanasinghe: Yes, this is related to the old text matching style behave tests.

tests/behaviour/steps/inspectcli.py
GDonRanasinghe: Yes, this is related to the old text matching style behave tests.

tests/behaviour/steps/inspectcli.py
GDonRanasinghe: We already have unit tests for this. If we are moving away from the test based approach for behave tests, we should test the objects? DSD components are objects.

src/csvcubed/models/sparqlresults.py
```diff
@@ -358,7 +358,7 @@ def _map_qube_component_sparql_result(
             str(result_dict["componentPropertyType"])
         ),
         csv_col_title=none_or_map(result_dict.get("csvColumnTitle"), str) or "",
-        observation_value_column_titles="",
+        observation_value_column_titles="", # This value is popuated after runnning the SELECT_OBS_VAL_FOR_DSD_COMPONENT_PROPERTIES sparql query. The _map_obs_val_for_dsd_component_properties_results will set the value for this property.
```
robons, line 361: Do you have the black formatter correctly installed and functioning here? It shouldn't allow a line this long.

src/csvcubed/utils/sparql_handler/sparqlmanager.py
```diff
@@ -222,7 +210,7 @@ def select_csvw_dsd_dataset_label_and_dsd_def_uri(
 
 
 def select_csvw_dsd_qube_components(
-    csvw_shape: Optional[CSVWShape],
+    cube_shape: Optional[CubeShape],
```
robons, line 213: I don't think the cube shape should really be optional at this point, it should require a value.

tests/unit/utils/sparqlhandler/test_sparqlmanager.py
```diff
@@ -240,6 +239,32 @@ def test_select_csvw_catalog_metadata_for_codelist():
     assert result.contact_point == "None"
     assert result.identifier == "Alcohol Content"
 
+def test_select_csvw_dsd_dataset_for_old_style_data_set():
```
robons, line 242: Can you make it clearer exactly what this test is doing. An `old style` data set only makes sense with the current context we have. It's going to be very difficult to understand what an old style data set is when a developer comes back and tries to understand this in future.

tests/unit/utils/sparqlhandler/test_sparqlmanager.py
```diff
@@ -240,6 +239,32 @@ def test_select_csvw_catalog_metadata_for_codelist():
...
+        csvw_metadata_rdf_graph
+    )
+    component_result: QubeComponentsResult = select_csvw_dsd_qube_components(
+        CubeShape.Standard, csvw_metadata_rdf_graph, result.dsd_uri, csvw_metadata_json_path
+    )
+    components = component_result.qube_components
+
+    assert result.dataset_label == "Alcohol Bulletin"
+    assert len(components) == 17
+    assert (
```
robons, line 260: As we've discussed previously, can you make sure you're not relying on the potentially random order in which the components come back?

robons (approved):
I'm approving this PR so we can get it out the way, but can you take a look at the new comments I've left and address them in your next PR?

Issue #626, opened 2022-10-25
Depreciation is coming (pandas in_place rename of categories is going)


canwaf:
```
D:\a\csvcubed\csvcubed\src\csvcubed\utils\qb\standardise.py:26: FutureWarning: Setting categories in-place is deprecated and will raise in a future version. Use rename_categories instead.
````

TBH all in_place stuff in pandas should be avoided, as it's all going away eventually.

- [x] Use the newer [rename_categories](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.cat.rename_categories.html) function as advised.
- [x] Check there are one or more tests which ensure the cateogrical rename functionality works and that they continue to function.
- [x] Manually verify that the warning message doesn't come up any longer. 

See [here](https://github.com/GSS-Cogs/csvcubed/blob/4a846b768329c843a5917b5bfcff3bbd69e2a64b/src/csvcubed/utils/qb/standardise.py#L216) for where the code change needs to be made. There *may* be other locations - double check.

PR #644, opened 2022-11-21
Rename pandas categories without inviting deprecation warning

robons:
Renaming categories on a pandas datacolumn in an in-place fashion yields a deprecation warning which isn't helpful for external users to see. 

This PR addresses #626 by avoiding in-place renaming of categories.

I also took the opportunity to remove some inplace replacements running in unrelated unit tests. Future versions of pandas will not support in-place replacement at all.


GDonRanasinghe (approved):
LGTM (looks good to me).

PR #654, opened 2022-12-01
Ensuring we don't publish documentation for pre-release versions

nimshi89:
Added in the extra check for Pre-release tag(-rc before the last digit in the tag).


robons (approved):
Merge this one in, then do a release candidate and make sure it *doesn't* release the docs. Then do a full release and make sure that it *does* release the docs.

PR #666, opened 2022-12-07
#662 required attributes warning for pivoted shape

GDonRanasinghe:
Satisfied #662

### DIFF
--- a/src/csvcubed/writers/qbwriter.py
+++ b/src/csvcubed/writers/qbwriter.py
@@ -16,7 +16,7 @@ import pandas as pd
 from csvcubed.models.cube.columns import CsvColumn, SuppressedCsvColumn
 from csvcubed.models.cube.cube import QbCube
 from csvcubed.models.cube.qb.columns import QbColumn
-from csvcubed.models.cube.qb.components.attribute import QbAttribute, QbAttributeLiteral
+from csvcubed.models.cube.qb.components.attribute import QbAttribute, QbAttributeLiteral, ExistingQbAttribute, NewQbAttribute
 from csvcubed.models.cube.qb.components.codelist import (
     NewQbCodeListInCsvW,
     NewQbCodeList,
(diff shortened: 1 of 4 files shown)


src/csvcubed/writers/qbwriter.py
```diff
@@ -540,19 +540,19 @@ def _get_primary_key_columns(self) -> List[str]:
...
-                # since we permit missing observation values where an `obsStatus` explains the reason.
-            )
-        )
+        if isinstance(column.structural_definition,(QbDimension, QbMultiUnits, QbMultiMeasureDimension)):
+            return True
+        elif (isinstance(column.structural_definition, QbAttribute) and column.structural_definition.get_is_required()):
+            if self.cube.is_pivoted_shape:
+                #If the cube is in pivoted shape, the attribute columns cannot be set to required.
+                _logger.warning(
+                    "Setting the column '%s' to not required as the cube is in pivoted shape.", column.csv_column_title
```
robons, line 549: Can you alter this message to say something like `Attribute column '%s' was marked as required, but the cube has multiple observed values columns. Attributes in multi-measure pivoted cubes cannot currently be marked as required.`

src/csvcubed/writers/qbwriter.py
```diff
@@ -540,19 +540,19 @@ def _get_primary_key_columns(self) -> List[str]:
...
-                isinstance(column.structural_definition, QbObservationValue)
-                and len(get_observation_status_columns(self.cube)) == 0
-                # We cannot mark an observation value column as `required` if there are `obsStatus` columns defined
-                # since we permit missing observation values where an `obsStatus` explains the reason.
-            )
-        )
+        if isinstance(column.structural_definition,(QbDimension, QbMultiUnits, QbMultiMeasureDimension)):
+            return True
+        elif (isinstance(column.structural_definition, QbAttribute) and column.structural_definition.get_is_required()):
+            if self.cube.is_pivoted_shape:
```
robons, line 546: It should be possible to have a required attribute in a single measure pivoted data set. We need to check whether it's pivoted **and** it's a multi-obs-val data set in this condtion here.

tests/unit/writers/qbwriter/test_qbwriter.py
```diff
@@ -199,7 +200,10 @@ def test_csv_col_required():
         csv_col = empty_qbwriter._generate_csvw_column_definition(col)
         required = csv_col["required"]
         assert isinstance(required, bool)
-        assert required
+        if isinstance(col.structural_definition, QbAttribute):
+            assert required == False
```
robons, line 204: This should only happen in a cube with multiple obs val columns. This cube has only one obs-val column. And perhaps more concerningly, this isn't even a pivoted shape cube.

robons (changes requested):

tests/unit/writers/qbwriter/test_qbwriter.py
nimshi89: Robsays hi

nimshi89 (commented):
I am smart.

robons (approved):
I approve.

Issue #669, opened 2022-12-15
Implement local retrieval of files if HTTP fails


CharlesRendle:
Following on from the spike (#401) this ticket aims to add a custom adapter to the cachedsession which will recognise when an http or https request has failed or produced a 'not ok' status code (400 - 599).
Upon this recognition, we should log an appropriate warning to the user and attempt to use the provided URL to find and retrieve a locally stored copy.

An exception to this adapter's use is during the schema validation in which we have no control over the get request made by the jsonschema library. The best we can do is log an error and proceed anyway when a RefResolutionError is returned.

ToDo:

- [ ] Write custom adapter and mount it to the session
- [ ] Write the logic which will catch faulty responses
    - [ ] Return the locally stored version of the file using the URL
- [ ] Ensure appropriate warnings are logged
- [ ] Implement a catch when a RefResolutionError is returned from schema validation
    - [ ] Log an error in this instance
- [ ] Unit tests for both types of http(s) responses and the ability to catch the schema validation error appropriately while continuing with the build process.

Issue #401, opened 2022-04-07
Use local copy of files if HTTP fails


canwaf:
This issue was raised due to a user of csvcubed being unable to access certain web resources due to them being behind a proxy server which white/black-lists various domains. This meant a user had an internet connection but wasn't able to load the resources that csvcubed was trying to load.

- If any of the HTTP request for the following files fails (for any reason, i.e. non-2xx HTTP status code) then our function should return the contents of the file from disk as if the HTTP request had succeeded (with an additional warning log printed in the log).
  - [ ] qube-config.json SCHEMA
  - [ ] code-list-config.json SCHEMA
  - [ ] qube-config templates

There are a few ways this could be implemented:

- Using the requests cache functionality we have defined in [cache.py](https://github.com/GSS-Cogs/csvcubed/blob/main/src/csvcubed/utils/cache.py), we could write the local files to the cache for use if the external HTTP request fails. This might not be possible.
- If the above isn't possible, we could also use a python request mocking library so long as it attempts to fetch the remote content before falling back on the local content.
- If the above isn't possible, we can simply update the three functions with logic to check for HTTP request failure and manually serve up the contents for the specific file requested.

If you don't managed to complete this within 8 points of effort, note down what the best approach is and create  new ticket to implement it.

robons:
https://findwork.dev/blog/advanced-usage-python-requests-timeouts-retries-hooks/

Hooks on the session object (the cached session) may be a suitable location to implement this.

PR #671, opened 2022-12-15
401 offline http

CharlesRendle:
This PR addresses the functionality investigated in ticket #401 and subsequently ticket #669 to provide further support for offline users.

Adds:
Retrieval of cube/codelist config documents and templates when an http(s) request fails (due to connectivity issues or faulty server responses) by mounting a custom adapter to the session.
There are unit tests for each of these scenarios narrowed down to the `load_json_document` and `_ get_properties_from_template_file` functions as well as demonstrating the scenario in which jsonschema returns a `RefResolutionError` because the GET request within that library will fail while offline.

Note for reviewer(s) - We have had to change the existing unit test `test_raise_error_works_for_none_existing_template_path()` in `tests/unit/readers/test_preset_column_templates.py ln:52` because a non existing template URL will trigger our new logic in the adapter and raise a `FileNotFoundError`. Please look at the assertion in this test as we were unsure what needed to change as the existing assertion passes.


src/csvcubed/utils/cache.py
```diff
@@ -1,4 +1,197 @@
...
+def generate_path_to_local_file(request_url: str) -> Path:
+    trimmed_url = str(request_url).removeprefix("https:")
+    if request_url[len(request_url) - 1] == "/":
+        path_to_local_file = map_url_to_file_path.get(
+            trimmed_url[: len(trimmed_url) - 1]
+        )
+    else:
+        path_to_local_file = map_url_to_file_path.get(trimmed_url)
+
+    if path_to_local_file is None:
```
robons, line 159: Can you just return an `Optional[Path]` instead of raising an exception?

src/csvcubed/utils/cache.py
```diff
@@ -1,4 +1,197 @@
...
+
+        return response
+
+    def close(self) -> None:
+        self.http_adapter.close()
+
+
+def generate_path_to_local_file(request_url: str) -> Path:
+    trimmed_url = str(request_url).removeprefix("https:")
+    if request_url[len(request_url) - 1] == "/":
```
robons, line 152: You've used `removeprefix`, why don't you use `removesuffix` for the `/` as well?

src/csvcubed/utils/cache.py
```diff
@@ -1,4 +1,197 @@
...
+
+
+        return response
+
+    def close(self) -> None:
+        self.http_adapter.close()
+
+
+def generate_path_to_local_file(request_url: str) -> Path:
+    trimmed_url = str(request_url).removeprefix("https:")
```
robons, line 151: Will this handle `http://` URLs as well?

src/csvcubed/utils/cache.py
```diff
@@ -1,4 +1,197 @@
...
+                )
+
+
+        return response
+
+    def close(self) -> None:
+        self.http_adapter.close()
+
+
+def generate_path_to_local_file(request_url: str) -> Path:
```
robons, line 150: Can this be made into a private function?

src/csvcubed/utils/cache.py
```diff
@@ -1,4 +1,197 @@
...
+    else:
+        path_to_local_file = map_url_to_file_path.get(trimmed_url)
+
+    if path_to_local_file is None:
+        raise KeyError
+
+    return path_to_local_file
+
+
+def create_local_copy_response(
```
robons, line 165: Can this be a private function too?

src/csvcubed/utils/cache.py
```diff
@@ -1,4 +1,197 @@
...
+
+    if response is not None:
+        successful_response.history = [response]
+
+    successful_response.reason = "OK"
+    successful_response.request = request
+
+    return successful_response
+
+
```
robons, line 194: Can we move all of this functionality out into a separate file to keep this `cache` file as clean as possible?

src/csvcubed/utils/json.py
```diff
@@ -45,15 +45,11 @@ def load_json_document(file_uri_or_path: Union[str, Path]) -> Dict[str, Any]:
...
-            if not http_response.ok:
-                raise Exception(
-                    f"Error loading JSON from URL '{file_uri_or_path}'. HTTP response: {http_response}."
-                )
 
             try:
                 return http_response.json()
             except Exception as e:
-                raise Exception(
+                raise Warning(
```
robons, line 52: Is this a real thing?

src/csvcubed/utils/json.py
```diff
@@ -45,15 +45,11 @@ def load_json_document(file_uri_or_path: Union[str, Path]) -> Dict[str, Any]:
             # Treat it as a URL
             _logger.debug("Loading JSON from URL %s", file_uri_or_path)
             http_response = session.get(file_uri_or_path)
-            if not http_response.ok:
```
robons, line 48: Is there a reason for removing this?

tests/unit/conftest.py
```diff
@@ -23,3 +25,39 @@ def configure_logging():
...
+def dummy_mapped_url():
+    """
+    This fixture is used to enable some tests to pass "bad input" URLs without causing an error
+    due to a corresponding local file path not existing. It maps the URLs used in those tests to
+    a file path that is known to exist. This allows connection errors and other such exceptions
+    to happen in a testing scenario.
+    """
+    # Add test URL to dictionary when ready to use fixture
+    test_dictionary = {
+        "//thisinputurlwillcauseanerror.com": APP_ROOT_DIR_PATH
```
robons, line 40: Could you stick comments here explaining the expected HTTP request behaviour when each of the following URLs are requested?

tests/unit/conftest.py
```diff
@@ -23,3 +25,39 @@ def configure_logging():
...
+        / "v1_3"
+        / "schema.json",
+        "//raw.githubusercontent.com/GSS-Cogs/csvcubed/main/src/csvcubed/readers/cubeconfig/v1_0/templates/calendar-hourx.json": APP_ROOT_DIR_PATH
+        / "readers"
+        / "cubeconfig"
+        / "v1_0"
+        / "templates"
+        / "calendar-year.json",
+    }
+    map_url_to_file_path.update(test_dictionary)
```
robons, line 57: Can you put the original values back in the dictionary after the test has completed?

tests/unit/utils/test_json.py
```diff
@@ -42,5 +45,72 @@ def test_loading_json_from_url():
         assert isinstance(document.get("columns"), list)
 
 
+def test_load_local_when_http_request_fails(dummy_mapped_url):
```
robons, line 48: I think this functionality should really be in a test file associated with the module we're really testing (in this case it's currently `cache.py`).

tests/unit/utils/validators/test_schema.py
```diff
@@ -151,5 +151,27 @@ def test_json_path_quote_escape():
...
+                "$ref": "http://thisisaurltoadocument/that/does/not/exist.json",
+            },
+        },
+    }
+    data = {
+        "publisher": "https://www.gov.uk/government/organisations/youth-justice-board-for-england-and-wales"
+    }
+
+    json_validation_errors = validate_dict_against_schema(data, schema)
+    assert len(json_validation_errors) == 0
```
robons, line 173: Do we want to test that we get a warning log message here too?

robons (commented):
Good work both of you, this has been a tricky task, but I think we're almost there.

I've left a few comments/suggestions which we can discuss.

-------

I've also got an approach to testing that it'd be worthwhile us trying...

I wonder whether it's possible to use vcrpy to 'properly' test this offline network functionality. i.e. we could explicitly provide a cassette which doesn't contain any of the requests it's necessary for the test to have (and by default vcrpy doesn't allow HTTP requests out for things it doesn't know about). 

In this way you should get some kind of connection failure which could be recovered using the functionality you've written. Given what we really want to ensure is that our application continues working (to the best of its ability) when offline, this would be the best test we could write.

It might be worth adding an extra ticket to investigate doing this.

src/csvcubed/utils/cache.py
```diff
@@ -1,4 +1,197 @@
...
+import requests
+from requests.adapters import BaseAdapter, HTTPAdapter
 from requests_cache import CachedSession
 
 
+from csvcubed.definitions import APP_ROOT_DIR_PATH
+
+_logger = logging.getLogger(__name__)
+
+_hard_codes_map_url_to_file_path = {
```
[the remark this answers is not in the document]
CharlesRendle: Changed name

src/csvcubed/utils/cache.py
```diff
@@ -1,4 +1,197 @@
...
+    / "schema"
+    / "codelist-config"
+    / "v1_1"
+    / "schema.json",
+    "//purl.org/csv-cubed/code-list-config/v1": APP_ROOT_DIR_PATH
+    / "schema"
+    / "codelist-config"
+    / "v1_1"
+    / "schema.json",  # v1 defaults to latest minor version of v1.*.
+}
```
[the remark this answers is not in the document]
CharlesRendle: Updated instructions

src/csvcubed/utils/cache.py
```diff
@@ -1,4 +1,197 @@
...
+
+
+def _get_url_to_file_path_map() -> Dict[str, Path]:
+    """
+    Creates a dictionary that maps URLs requested by load_json_document to their corresponding local storage files,
```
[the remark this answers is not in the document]
CharlesRendle: Changed description

src/csvcubed/utils/cache.py
```diff
@@ -1,4 +1,197 @@
...
+        )
+        map_uri_to_file_path[uri] = template_file
+
+    return map_uri_to_file_path
+
+
+map_url_to_file_path = _get_url_to_file_path_map()
+
+
+class CustomAdapterServeSomeFilesLocally(BaseAdapter):
```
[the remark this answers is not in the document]
CharlesRendle: Changed adapter title and added a docstring description

src/csvcubed/utils/cache.py
```diff
@@ -1,4 +1,197 @@
...
+                stream=stream,
+                timeout=timeout,
+                verify=verify,
+                cert=cert,
+                proxies=proxies,
+            )
+
+        except requests.exceptions.ConnectionError as e:
+            try:
+                _logger.debug("The connection error has been found")
```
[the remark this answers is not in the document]
CharlesRendle: Added e to debug_log_exception and a comment.

src/csvcubed/utils/cache.py
```diff
@@ -1,4 +1,197 @@
...
+
+                path_to_local_file = generate_path_to_local_file(request.url)
+
+                _logger.debug("The local file path is: %s", path_to_local_file)
+
+                return create_local_copy_response(path_to_local_file, request)
+
+            except KeyError:
```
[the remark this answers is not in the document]
CharlesRendle: Removed try/excepts on KeyError. Now raises the exception if path_to_local_file is None.

src/csvcubed/utils/cache.py
```diff
@@ -1,4 +1,197 @@
...
+                    f"URL {request.url} did not produce a response and a local copy could not be found at the corresponding mapped path."
+                ) from e
+
+        if response.status_code >= 400 and response.status_code < 600:
+            _logger.info(f"The status code is {response.status_code}")
+            try:
+                path_to_local_file = generate_path_to_local_file(request.url)
+                return create_local_copy_response(path_to_local_file, request, response)
+
+            except KeyError:
```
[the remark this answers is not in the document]
CharlesRendle: Addressed above.

src/csvcubed/utils/cache.py
CharlesRendle: Changed to return Optional[Path] and removed the if statement.

src/csvcubed/utils/cache.py
CharlesRendle: added a remove suffix in the event of `/`

src/csvcubed/utils/cache.py
CharlesRendle: added a removeprefix for `http:` and removed the logic to handle the trailing forward slash

src/csvcubed/utils/json.py
robons: Don't forget about this one.

tests/unit/conftest.py
robons: Don't forget about this one.

src/csvcubed/utils/createlocalcopyresponse.py
```diff
@@ -0,0 +1,203 @@
...
+    successful_response.url = path_to_local_file.as_uri()
+
+    successful_response.encoding = "utf-8"
+
+    if response is not None:
+        successful_response.history = [response]
+
+    successful_response.reason = "OK"
+    successful_response.request = request
+    successful_response.ok
```
robons, line 201: You've left this in here.

PR #678, opened 2022-12-21
663 getting unit label for unit uri

GDonRanasinghe:


robons:
Replaced by #679 which has the appropriate changes in it without random other ones.

Issue #642, opened 2022-11-18
Improve the peformance  of `_generate_printables`


nimshi89:
The `select_csvw_dsd_qube_components` SPARQL query take up the majority of this time - 59% of the whole inspect command runtime. This SPARQL query should be split up into smaller composable SPARQL queries where results are cached against the `SparqlManager`. By doing this we can avoid the unnecessarily large left joins that make this large SPARQL query so poorly performant. N.B. The left joins are generated by the OPTIONAL parts of the SPARQL queries.

- [x] Move `select_dsd_qube_components` to the DataCubeState class.
- [x] Split out the top part of `select_dsd_qube_components` query and move it to the DataCubeState class.
- [x] Fix any breaking unit and behave tests for the above.
 
- [x] Rename `SparqlManager` to `SparqlQueryManager` to make it clearer what its purpose is.
- [ ] Taking the [select_dsd_qube_components query](https://github.com/GSS-Cogs/csvcubed/blob/pivoted-shape-feature/src/csvcubed/utils/sparql_handler/sparql_queries/select_dsd_qube_components.sparql) experiment with breaking the query into smaller parts which could achieve the same goal and test query execution times to prove to yourself that this will improve overall performance against the existing SPARQL query.
- [ ] If you have enough time in this ticket, iImplement these changes into the codebase, else create a follow up ticket explaining exactly what implementation approach to take.
- [ ] Add appropriate unit tests for each of the new SPARQL queries thart have been added.

When implementing the new SPARQL queries you should ensure that the results are available as [cached properties](https://docs.python.org/3/library/functools.html#functools.cached_property) on the `SparqlQueryManager`. This means they should only be run once, and only executed when necessary (lazy-loading). Follow the pattern established in #611.

Rob's suggestions for how we could break this query up into composable components:

* Have a single query which, lists the DSD components and their properties. Ideally this could return what the type of component is (dimension/measure/etc.) and, if an attribute, whether the attribute is required or not (all other components are required). We should also bring back the component order as well as the `componentPropertyLabel` (if it exists). The resulting model should be a map of DSD URI -> the set of component properties.
[...]

PR #684, opened 2022-12-30
#642 Speeding up Inspect API's Data Structure Definition SPARQL Query

robons:
Initial informal tests suggest an approx 25% improvement in performance.

@nimshi89 and myself will be doing more thorough testing of this using the stress test, and then review profiler results - but it'd be good to get a review going on this ASAP.

Done as part of #642


src/csvcubed/cli/inspect/inspectdatasetmanager.py
```diff
@@ -99,13 +101,15 @@ def get_standard_shape_measure_col_name_from_dsd(
         ComponentPropertyAttributeURI.MeasureType.value,
     )
 
-    if len(filtered_components) == 0 or filtered_components[0].csv_col_title == "":
+    if len(filtered_components) == 0 or filtered_components[0].used_in_columns == "":
```
robons, line 104: This is wrong.

src/csvcubed/cli/inspect/inspectdatasetmanager.py
```diff
@@ -124,13 +128,15 @@ def get_standard_shape_unit_col_name_from_dsd(
         ComponentPropertyAttributeURI.UnitMeasure.value,
     )
 
-    if len(filtered_components) == 0 or filtered_components[0].csv_col_title == "":
+    if len(filtered_components) == 0 or filtered_components[0].used_in_columns == "":
```
robons, line 131: This won't work.

src/csvcubed/models/sparqlresults.py
```diff
@@ -438,75 +449,77 @@ def _map_qube_component_sparql_result(
...
+        csv_column_definitions = map_csv_url_to_column_definitions[csv_url]
+
+        measure_uris = {
+            c.property for c in components if c.property_type == str(QB.MeasureProperty)
+        }
+
+        observed_value_columns = [
+            c
+            for c in csv_column_definitions
+            if (not c.virtual) and c.property_url in measure_uris
```
robons, line 490: I don't think this matching is working at the moment. When debugging, I saw he `observed_value_columns` empty, when it needs to have some values. I think we need to do `startsWith` checking or similar.

src/csvcubed/models/sparqlresults.py
```diff
@@ -438,75 +449,77 @@ def _map_qube_component_sparql_result(
...
             ]
 
-            if len(obs_val_col_title_result_for_component) == 1:
-                dsd_component_result.observation_value_column_titles = (
-                    obs_val_col_title_result_for_component[
-                        0
-                    ].observation_value_column_titles
-                )
+            component.used_in_columns = [
+                c for c in columns_using_this_component if c.title is not None
```
robons, line 501: We should get rid of the `if c.title is not None` off the end here.

src/csvcubed/models/sparqlresults.py
```diff
@@ -438,75 +449,77 @@ def _map_qube_component_sparql_result(
...
+                c.about_url
+                for c in columns_using_this_component
+                if c.about_url is not None
+            }
+
+            component.used_by_observed_value_columns = [
+                c
+                for c in observed_value_columns
+                if c.about_url in columns_using_this_component_about_urls
+                and c.title is not None
```
robons, line 514: We should get rid of the `and c.title is not None` off the end here.

src/csvcubed/utils/csvdataset.py
```diff
@@ -244,7 +248,7 @@ def transform_dataset_to_canonical_shape(
         obs_val_col_titles_about_urls = (
             data_cube_state.get_obs_val_col_titles_about_urls_for_csv(csv_url)
         )
-        col_names_col_titles = data_cube_state.get_col_name_col_title_for_csv(csv_url)
+        col_names_col_titles = data_cube_state.get_column_definitions_for_csv(csv_url)
```
robons, line 251: We need to rename this variable.

tests/unit/utils/test_csvdataset.py
```diff
@@ -419,13 +419,13 @@
 ]
 
 _col_names_col_titles = [
```
robons, line 421: I'm sure this'll need to change quite a bit.

src/csvcubed/models/sparqlresults.py
```diff
@@ -438,75 +449,77 @@ def _map_qube_component_sparql_result(
         property_type=get_component_property_type(
             str(result_dict["componentPropertyType"])
         ),
-        csv_col_title=none_or_map(result_dict.get("csvColumnTitle"), str) or "",
-        observation_value_column_titles="",  # This value is popuated after runnning the SELECT_OBS_VAL_FOR_DSD_COMPONENT_PROPERTIES sparql query. The _map_obs_val_for_dsd_component_properties_results will set the value for this property.
+        used_in_columns=none_or_map(result_dict.get("csvColumnTitle"), str) or "",
```
robons, line 452: This is wrong.

src/csvcubed/cli/inspect/inspectdatasetmanager.py
nimshi89: that means on line 108 "return filtered_components[0].csv_col_title" has to be changed as well right?

GDonRanasinghe (approved):
This looks good to me Rob - nice job!

PR #685, opened 2023-01-06
Muazzam chaud/issue527

MuazzamChaud:


robons (approved):

.github/workflows/reusable-release_gssutils.yaml
```diff
@@ -0,0 +1,14 @@
+name: Trigger a workflow at gssutils that updates csvcubed package
```
robons, line 1: Might be nice to make this name a bit shorter, perhaps `Trigger csvcubed update in gss-utils`?

Issue #632, opened 2022-11-08
Investigate performance improvements for Github Actions build


robons:
The csvcubed build process is currently very slow on GitHub Actions due to these being a single build process for Windows and Linux which needs to install tools every time the process is run. This process could be improved by creating some reusable docker images which are pre-built with all of the tools necessary.

Unfortunately, linux docker containers cannot run in the Windows environment, so we will need to create a docker container for the linux environment and another one for the Windows environment. We either need to have a separate docker image built for every version of python we want to test, or we need to be able to have multiple versions of python installed within each container (think pyenv) which we can switch between at runtime. 

The first step of this task should be to progress towards the Windows docker container(s) (since the Windows part of the testing process is always the slowest). Make sure the container has all necessary versions of python installed as well as csvlint, csv2rdf and the sparql tests. 

I recommend you do this in a Windows virtual machine on your local macbook - do not try and do this in GitHub actions just yet as that is a terrible way to trial + debug.

- [ ] Once built, publish the container manually to docker hub so we can use it in the build pipeline.
- [ ] Once complete, create a new task to make use of the windows docker image within the csvcubed github actions pipeline.
- [ ] Once complete, create another new task to automate the build and deployment of this windows docker image using GitHub actions from a github repository.

![image.png](https://images.zenhubusercontent.com/60d35dd0922789587fc8922d/0a93350a-7075-4057-9b8a-d4fb689e169e)

The windows tests take >13mins and the  linux ones take >6 mins. Fix the problems with Windows performance first.

nimshi89:
https://console.cloud.google.com/marketplace/product/cloud-infrastructure-services/docker-windows-2019?project=optimum-bonbon-257411

SarahJohnsonONS:
Initially attempted using full Windows Server image - this took approximately 6 mins to build the container, which would not improve the performance of the test_in_environments action for Windows builds. Switched to Nanoserver image, but this introduced issues with running commands as Powershell is not available by default, and paths were not being persisted between commands. 

There may be some value in exploring increased caching of steps in the test_in_environments action, e.g. for installing Python, Pyright, Poetry - see the screenshot below for timings of these steps.
![unnamed.png](https://images.zenhubusercontent.com/633c305c3267484177e9bc88/612a4331-e8bc-4f81-a4c6-e07adda77636)

PR #692, opened 2023-01-18
#632 Trying Windows Docker container for tests

SarahJohnsonONS:


robons:
Closing this PR since we're closing the ticket due to it being unnecessarily difficult to create viable Windows docker containers which would speed up the build process. The only workable docker container was ~3.3GB in size and took 6 mins to download and run from docker hub. This wouldn't have improved the runtime of our builds in any significant way and may have increased the time it took to run them.

Issue #697, opened 2023-01-19
Incorrect linking of slice to observations in pivoted shape


robons:
Given a multi-measure pivoted data set, when we generate the CSV-W we are using the wrong predicate to link the row's Slice to each of the individual Observation values within the slice. 

We currently link the slice to the observations as per the following example:

```ttl
eg:slice2 a qb:Slice;
    qb:Observation eg:o1b, eg:o2b, eg:o3b, ... .
```

But the correct way to link a slice to its observations is (see [here](https://www.w3.org/TR/vocab-data-cube/#h3_slices)):

```ttl
eg:slice2 a qb:Slice;
    qb:observation eg:o1b, eg:o2b, eg:o3b, ... .
```

Note that we're using `qb:Observation` in the first example whereas it should use the lowercase predicate `qb:observation`.

Example dataset containing problem: [Official reported estimates of the percentage of the population testing positive for COVID-19, England](https://staging.gss-data.org.uk/cube/explore?uri=https%3A%2F%2Fraw.githubusercontent.com%2Fcanwaf%2Fprovider-covid-infection-survey%2FworkingCopy%2Fout%2Fofficial-reported-estimates-of-the-percentage-of-the-population-testing-positive-for-covid-19-england.csv%23dataset-catalog-entry).

- [ ] Make sure we have a behaviour test to ensure we're expressing it correctly. There may already be one, but if not create one.

## Triage

This currently doesn't strictly have an impact given that none of the tools we use make use of the slice we generate (it's there simply becasue we need a Resource with a URI to be serialised for each row in the pivoted data shape which links to each of the observations on the row - else they don't get written to RDF). However, this bug does affect all of the CSV-Ws we're generating using csvcubed. 

Probability: Very High, Impact: Very Low (this may be raised when we start needing to read these triples)

Combined score: 5 = Important.

PR #708, opened 2023-01-25
697 Incorrect linking of slices to observations in pivoted shape

NickPapONS:
This was a quick fix to qbwriter.py where in a pivoted multi measure dataset, we were linking a slice to the corresponding observation values using the wrong predicate. This simply meant changing the predicate in this virtual column creation from qb:Observation to qb:observation, then adjusting the relevant tests so they expect the correct results when a CSV-W is generated.

Statisfied #697.

### DIFF
--- a/tests/behaviour/qbwriter.feature
+++ b/tests/behaviour/qbwriter.feature
@@ -66,27 +66,27 @@ Feature: Test outputting CSV-Ws with Qb flavouring.
     And csvlint validation of all CSV-Ws should succeed
     And csv2rdf on all CSV-Ws should succeed
     And the RDF should contain
-    """
+      """
       @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
       @prefix qudt: <http://qudt.org/schema/qudt/>.
       @prefix om2: <http://www.ontology-of-units-of-measure.org/resource/om-2/>.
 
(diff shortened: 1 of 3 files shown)


canwaf (approved):
Yes.

tests/behaviour/qbwriter.feature
```diff
@@ -66,53 +66,53 @@ Feature: Test outputting CSV-Ws with Qb flavouring.
...
       <{{rdf_input_directory}}/some-qube.csv#unit/some-extending-unit>
-        a qudt:Unit, om2:Unit;
-        qudt:isScalingOf <{{rdf_input_directory}}/some-qube.csv#unit/some-base-unit>;
-        qudt:hasQuantityKind <http://some-quantity-kind>;
-        qudt:conversionMultiplier "25.123123"^^xsd:float;
-        om2:hasUnit <{{rdf_input_directory}}/some-qube.csv#unit/some-base-unit>;
-        om2:hasFactor "1000.0"^^xsd:float;
-        rdfs:label "Some Extending Unit"@en.
-    """
+      a qudt:Unit, om2:Unit;
```
robons, line 108: Can you try not to alter the indentation in future? It makes it harder to see the pattern without it.

robons (approved):

Issue #696, opened 2023-01-19
Inspect api populate data_cube_state.py to use cached properties


NickPapONS:
This is a follow-up to #652, which involved performing some required refactoring and changing the python code focused around the inspect SPARQL queries to enable use of the cached_property feature, allowing cached results to remain available and be accessed more easily.

The DataCubeState class (defined in data_cube_state.py) has already been populated with some cached properties, and these can be used somewhat as a reference for implementing this for the several remaining queries.

However, each query can be very different to the rest and will require reading and understanding it and its relevant code to plan how to create this cached_property, as well as considering if its existing code may need some refactoring/alterations to allow the cached_property function to be created.

The rough requirements for each query are:

-  A cached_property function that caches/accesses the results of the SPARQL query.
- An API function (these are written in the same file, below the cached properties) which gets the result from the cached property function.
- A suitable unit test.
- After this is complete, it may be best to verify that all the other unit/behaviour tests still pass. The changes to the sparql queries and associated code may cause failures in existing tests which will need to be adjusted to accommodate the new changes.

A list of the queries that require cached property functionality, as well as some additional information to show which ones specifically belong to data_cube_state, can be found here: https://docs.google.com/spreadsheets/d/1mzHuDEpPKLrCPRXL0dj9bznWp3I9ZRZZ4kRKULhhVkg/edit#gid=0

The queries which need to be moved over are:

- [x] select_codelists_and_cols.sparql
- [x] select_cols_w_suppress_output.sparql - This can be replaced by the existing generic Columns cached property which has this information on it (delete SPARQL query)
- [ ] select_csvw_table_schema_file_dependencies.sparql
- [x] select_dsd_datasetlabel_and_uri.sparql
- [x] select_dsd_qube_components.sparql
- [x] select_is_pivoted_shape_for_measures_in_data_set.sparql
- [ ] select_obs_val_for_dsd_component_properties.sparql - This can be extracted from the DSD components property (delete SPARQL query)
- [ ] select_single_unit_from_dsd.sparql - This can be extracted from the DSD components property (delete SPARQL query)

For the queries which are to be deleted, we need to make sure to remove all unit tests and other code references to them (including results mapper functions). We then need to write `get_` API functions on the DataCubeState class to get the results that are necessary for the original callers of the query. 

- [x] Rename class (and file) to `DataCubeInspector`.

Issue #652, opened 2022-11-29
inspect command sparql results should use cached properties


GDonRanasinghe:
We are aiming to progress inspect command to behave like an API. To this end, we want to use the @cached_property python feature for inspect command results. The task is to update all the inspect command sparql results to use this feature. A reference is available in the data_cube_state.py class which is implemented as part of the pivoted shape work.

In addition, we may need to think about bringing `?dsd_uri` back out of the `select_unit_col_observed_col_title_and_about_url` sparql query as well and make a double dictionary - https://github.com/GSS-Cogs/csvcubed/pull/653/files#r1037033633

GDonRanasinghe:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Features-Squad-61b72275ac896c0010a1b00b/issues/GSS-Cogs/csvcubed/652?planning-poker) with Zenhub @CharlesRendle @NickPapONS @nimshi89 @SarahJohnsonONS

PR #723, opened 2023-02-03
#696 cached properties in data cube state

NickPapONS:
In this ticket, we made several changes to SPARQL queries that are relevant to the recently created data_cube_state.py file (now data_cube_inspector). This meant both altering a few queries slightly, as well as removing some entirely, then reworking their code implementations to use the cached property feature. For SPARQL queries that were removed, their functionality has been replaced by now accessing the information they were being used for from somewhere already available, for example; the query select_cols_w_suppress_output has been removed entirely, as the information it was being used to retrieve can now be retrieved from the column definitions property. This also happened with several queries that are now replaced by their information being available from the dsd_qube_components property. 

The remaining queries were refactored and reworked to make use of the cached_property feature, building upon the work that was started in #652 inspect command sparql results should use cached properties. This involved creating a cached_property function that stores the query's result, then a "getter" function that retrieved the stored results, then altering any existing places that called the query to instead get its result using this "getter". All the cached property functions in data_cube_inspector now have suitable unit tests.

Many places in csvcubed that used to call SPARQL queries directly now instead access their required information from the cached properties. This should hopefully lead to performance improvements (though we haven't measured this as part of this ticket), as SPARQL queries tend to be quite performance heavy and are now going to be run fewer times throughout the code base.

Note that one query listed in the ticket description, "select_csvw_table_schema_file_dependencies", turned out to not belong in data_cube_state, and therefore has been left untouched. We also doubt it would be needed in any other inspector/state file, as implementing a cached property for it would not actually reduce the number of times this query is called in the code base.


external-docs/docs/guides/configuration/qube-config.md
```diff
@@ -50,15 +50,7 @@ A column is assumed to be a dimension unless otherwise configured using the `typ
 }
 ```
 
-**If a column mapping is not defined in the `qube-config.json` file for a given CSV column, the column is [configured by convention](./convention.md).**  To ignore a column and to not configure it by convention, the column's definition to `false`. This will ensure the column will not be present in the CSV-W when built by csvcubed.
-
-```json
-{ ...
- "columns": {
-  "Suppressed column": false
```
robons, line 58: Noooo, this needs to come back. @canwaf added it a week or so again.

external-docs/docs/guides/configuration/templates.md
```diff
@@ -31,41 +31,31 @@ In the above example, we have reused the `year` template, but given the dimensio
 
 ## Available templates
 
-The templates are currently in version 1.0; with future development planned. You can see the entire list of the templates inline below or on [csvcubed github](https://github.com/GSS-Cogs/csvcubed/tree/main/src/csvcubed/readers/cubeconfig/v1_0/templates). The `from_template` functionality fetches the most recent version of our templates from the web.
+The templates are currently in version 1.0; with future development planned. You can see the entire list of the templates inline below or on [csvcubed github](https://github.com/GSS-Cogs/csvcubed/tree/main/src/csvcubed/readers/cubeconfig/v1_0/templates). The `from_template` funcitonality fetches the most recent version of our templates from the web.
```
robons, line 34: You've introduced a typo :P

I don't think you should be making any changes at all in this file in this PR.

src/csvcubed/cli/inspect/metadataprinter.py
```diff
@@ -150,24 +151,25 @@ def get_datacube_results(self):
...
-        csvw_state = self.state.csvw_state
+        assert isinstance(self.state, DataCubeInspector)  # Make pyright happier
 
         self.result_qube_components = self.state.get_dsd_qube_components_for_csv(
             self.primary_csv_url
         )
 
-        self.result_dataset_label_dsd_uri = (
-            select_csvw_dsd_dataset_label_and_dsd_def_uri(csvw_state.rdf_graph)
+        self.result_cube_table_identifiers = self.state.get_cube_identifiers_for_csv(
```
robons, line 160: I'd be tempted to call this `primary_cube_table_identifiers`

src/csvcubed/cli/inspect/metadataprinter.py
```diff
@@ -150,24 +151,25 @@ def get_datacube_results(self):
...
+            self.primary_csv_url
         )
-        self.result_cols_with_suppress_output_true = (
-            select_cols_where_suppress_output_is_true(csvw_state.rdf_graph)
+        self.result_column_definitions = self.state.get_column_definitions_for_csv(
```
robons, line 163: I'd call this `primary_csv_column_definitions` to avoid confusion over columns defined in any other CSVs referenced in the CSV-W.

I'd also do the same with `suppressed_columns` and `result_code_lists` below as well.

src/csvcubed/cli/inspect/metadataprinter.py
```diff
@@ -150,24 +151,25 @@ def get_datacube_results(self):
...
+            self.primary_csv_url
         )
-        self.result_code_lists = select_dsd_code_list_and_cols(
-            csvw_state.rdf_graph,
-            self.result_dataset_label_dsd_uri.dsd_uri,
-            csvw_state.csvw_json_path,
+
+        self.suppressed_columns = self.state.get_suppressed_columns_for_csv(
```
robons, line 167: I do wonder whether we need to store `suppressed_columns` separately since the information is already contained within `result_column_definintions`.

src/csvcubed/cli/inspect/metadataprinter.py
```diff
@@ -260,7 +262,8 @@ def dsd_info_printable(self) -> str:
 
         :return: `str` - user-friendly string which will be output to CLI.
         """
-        return f"- The {self.csvw_type_str} has the following data structure definition:{self.result_dataset_label_dsd_uri.output_str}{self.result_qube_components.output_str}{self.result_cols_with_suppress_output_true.output_str}"
+        # get_printable_list_str called directly here for suppressed columns - see comment above re alternative approaches.
```
robons, line 265: Can we remove this comment now?

src/csvcubed/models/sparqlresults.py
```diff
@@ -143,14 +115,21 @@ class CodelistsResult:
     Model to represent select codelists sparql query result.
     """
 
-    codelists: list[CodelistResult]
+    codelists: List[CodelistResult]
     num_codelists: int
 
     @property
     def output_str(self) -> str:
+        def alter_code_list_for_text_representation(code_list: CodelistResult) -> Dict:
```
robons, line 123: Since we're doing work here, can we get rid of the `output_str` property and move it into the MetadataPrinter (or other appropriate location)?

src/csvcubed/readers/cubeconfig/v1_0/templates/government-week.json
```diff
@@ -2,5 +2,5 @@
     "type": "dimension",
     "from_existing": "http://purl.org/linked-data/sdmx/2009/dimension#refPeriod",
     "label": "Government Week",
-    "cell_uri_template": "http://reference.data.gov.uk/id/government-week/{+<column_name>}"
-}
+    "cell_uri_template": "http://reference.data.gov.uk/id/goverment-week/{+<column_name>}"
```
robons, line 5: Don't undo this change please.

src/csvcubed/readers/cubeconfig/v1_0/templates/government-year.json
```diff
@@ -2,5 +2,5 @@
     "type": "dimension",
     "from_existing": "http://purl.org/linked-data/sdmx/2009/dimension#refPeriod",
     "label": "Goverment Year",
-    "cell_uri_template": "http://reference.data.gov.uk/id/government-year/{+<column_name>}"
-}
+    "cell_uri_template": "http://reference.data.gov.uk/id/goverment-year/{+<column_name>}"
```
robons, line 5: Don't undo this change please.

src/csvcubed/readers/cubeconfig/v1_0/templates/gregorian-interval.json
```diff
@@ -1,6 +0,0 @@
-{
```
robons, line 1: Don't delete this file please.

src/csvcubed/readers/cubeconfig/v1_0/templates/mixed-period.json
```diff
@@ -1,7 +0,0 @@
-{
```
robons, line 1: Don't delete this file please.

src/csvcubed/readers/cubeconfig/v1_0/templates/preset_column_config.json
```diff
@@ -12,10 +12,7 @@
     "government-half-year": "government-half-year.json",
```
robons, line 12: I don't think you should be making any changes in this file in this PR.

src/csvcubed/schema/cube-config/v1_4/schema.json
```diff
@@ -677,10 +677,7 @@
                         "government-half-year",
                         "government-quarter",
                         "government-week",
-                        "mixed-period",
-                        "gregorian-instant",
-                        "gregorian-interval",
-                        "statistical-geography"
+                        "gregorian-instant"
```
robons, line 680: I don't think you should be making any changes in this file in this PR.

src/csvcubed/utils/printable.py
```diff
@@ -23,23 +23,6 @@ def get_printable_list_str(items: List) -> str:
     return output_str
 
 
-def get_printable_tabular_list_str(items: List) -> str:
```
robons, line 26: Where has this been moved to?

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
```diff
@@ -113,20 +123,44 @@ def _detect_shape_for_cube(
...
-        """The function returns a single unit from the uri if there is any."""
+        """
+        Get a specific unit, by its uri.
+        """
         return self._units.get(uri)
 
     def get_units(self) -> List[UnitResult]:
-        """This function gets the value of the unit and returns it as a list."""
+        """
+        Get all units in the data cube.
```
robons, line 157: Well, technically there may be multiple data cubes in the graph.

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
```diff
@@ -151,14 +185,33 @@ def get_cube_identifiers_for_data_set(
...
         return get_from_dict_ensure_exists(self._cube_shapes, csv_url)
+
+    def get_code_lists_and_cols(self, csv_url: str) -> CodelistsResult:
+        """
+        Get the codelists and columns associated with the given csv url.
+        """
+
+        return self._codelists_and_cols.get(csv_url, CodelistsResult([], 0))
+
+    def get_suppressed_columns_for_csv(self, csv_url: str) -> List[str]:
```
robons, line 205: I'm not sure this really adds much over just retrieving the columns definitions yourself for the csv and doing some filtering.

src/csvcubed/utils/sparql_handler/sparql_queries/select_data_set_dsd_csv_url.sparql
```diff
@@ -1,13 +1,15 @@
 PREFIX qb: <http://purl.org/linked-data/cube#>
 PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
 PREFIX csvw: <http://www.w3.org/ns/csvw#>
 
-SELECT ?dataSet ?dsd ?csvUrl
+SELECT ?dataSet ?dsd ?dataSetLabel ?csvUrl
```
robons, line 6: Do we need this label in here anymore?

tests/unit/utils/sparqlhandler/test_data_cube_inspector.py
```diff
@@ -180,62 +225,6 @@ def test_get_dsd_qube_components_for_csv():
         True,
     )
 
-    component = get_dsd_component_by_property_url(
```
robons, line 183: Why has this section disappered?

external-docs/docs/guides/configuration/qube-config.md
NickPapONS: Will change it back, this must have somehow happened during the merge with main?

tests/unit/utils/sparqlhandler/test_data_cube_inspector.py
NickPapONS: I believe we still test this function in this file, but now we do it a bit more thoroughly by testing it on a standard shape input, single measure pivoted, multi measure pivoted inputs. That's why it looks like this test has disappeared. If this is unnecessary then I can undo the change and only have a single test.

NickPapONS: This unit test is gone because it has been split into three different tests now, testing the get_dsd_component_by_property_url and get_dsd_qube_components_for_csv functions with a different input (e.g. standard shape data cube, pivoted shape single measure, pivoted shape multi measure). I believe we received this change from the merge with main.

src/csvcubed/cli/inspect/metadataprinter.py
NickPapONS: I have left this as it is for now because I wanted to ask; does this mean we should get rid of the get_suppressed_columns_for_csv function? I agree we can definitely get what we need from the column definitions, just not sure if it is necessary or not.

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
NickPapONS: See previous comment on this function's use (will delete this function if answer is yes)

src/csvcubed/utils/sparql_handler/sparql_queries/select_data_set_dsd_csv_url.sparql
NickPapONS: This label is here because we took it from another query (named select_dsd_dataset_label_and_uri) then deleted that query. But if we don't need the label at all, we can remove it from here as well.

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
```diff
@@ -21,21 +23,29 @@
...
-class DataCubeState:
-    """Provides access to inspect the data cubes contained in an rdflib graph."""
-
+class DataCubeInspector:
     csvw_state: CsvWState
 
     # Private cached properties.
 
+    @cached_property
+    def _column_definitions(self) -> Dict[str, List[ColumnDefinition]]:
```
robons, line 42: Isn't this defined in the CsvWInspector/CsvWState class now?

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
```diff
@@ -113,20 +123,44 @@ def _detect_shape_for_cube(
...
+        return select_dsd_code_list_and_cols(
+            self.csvw_state.rdf_graph,
+            self.csvw_state.csvw_json_path,
+        )
+
+    """
+    Public getters for the cached properties.
+    """
+
+    def get_column_definitions_for_csv(self, csv_url: str) -> List[ColumnDefinition]:
```
robons, line 140: Isn't this defined in the CsvWInspector/CsvWState class now?

src/csvcubed/cli/inspect/metadataprinter.py
robons: I think we should get rid of that function since we can derive the same information from the column definitions.

robons (changes requested):
Looks pretty close to me, but if you could remove the `column` functions from the DataCubeInspector (since they should be in the CsvWInspector now), and ideally get rid of the function which only exists to return the name of suppressed columns, then I'd be happier.

It might be a good point to merge in the changes from your CsvWInspector ticket first.

robons (approved):
I thought I'd give you two a bit of a break from the back-and-forth and go and tidy up a few things myself.

Looks good now.

Issue #705, opened 2023-01-24
Move SPARQL queries to CsvWState


robons:
- [x] `select_table_schema_properties.sparql` - This needs to be moved to the CsvWState class as it can apply to any type of CSV-W. It may be that we need to add a `get_` API function in the CodeListInspector class to filter this information in a more appropriate way for the caller.
- [x] `select_codelist_primary_key_by_csv_url.sparql` - We may need to add a `get_` API function inside the `DataCubeInspector` class to get the information that the caller would want in a more natural way.
- [x] Write unit tests for these queries.
- [x] Rename the `CsvWState` class (and file) to `CsvWInspector`

PR #727, opened 2023-02-09
#705 move sparql queries to csvw state

NickPapONS:
This ticket is further work following up on the spike ticket #652 implementing the cached property feature to the inspect SPARQL queries. The queries relevant here are specifically those that do not belong either in data_cube_state or code_list_state, and so instead are placed in csvw_state, a more generic class that exists separately from the more specific data cube or code list state/inspector classes (and exists before either is initialized). This means these queries are focused around obtaining generic information from an input, likely before it is determined if they are data cubes or code lists.

Adding the cached property functionality to the two queries specified in the ticket was similar to how the process was done in other similar tickets. This meant adding a cached_property function that runs the query and stores the result, then a "getter" function that accesses this result. Then, any existing location in code that called the query functions directly was altered to instead use the "getter" to access its required information, meaning they no longer run the (often performance heavy) queries anymore. These new functions are also unit tested.

Satisfies ticket #705.

Update 17/2/23: All comments addressed. One failing unit test and one failing behaviour test which still need investigating:

cli.feature "The csvcubed build command should output validation errors file" - no validation-errors.json file is being created?

 test_skoscodelistreader: test_num_of_variables_in_about_url_invalid_exception - KeyError: "Could not find catalog metadata in primary graph 'file:///workspaces/csvcubed/tests/test-cases/readers/skoscodelistreader/about-url-invalid-variable.csv-metadata.json'."


src/csvcubed/utils/sparql_handler/csvw_inspector.py
```diff
@@ -6,17 +6,23 @@
...
+    select_primary_key_col_names_by_csv_url,
+    select_table_schema_properties,
 )
 
 
 @dataclass
-class CsvWState:
+class CsvWInspector:
     rdf_graph: rdflib.ConjunctiveGraph
     csvw_json_path: Path
```
SarahJohnsonONS, line 28: Should all cached properties be private? Both catalog_metadata and csvw_type are currently not set as private.

robons: Not necessarily. If they're to be directly accessed by a user outside the class they should be public. Since `csvw_type` is used directly (and not via a `get_` function - since this function would do nothing useful) by external callers (and not just tests), it should be public. 

The same will be true of `catalog_metadata` once we've written functions which map the cube/code-list URIs to the catalogue_metadata associated with them.

src/csvcubed/cli/inspect/metadataprinter.py
```diff
@@ -203,13 +203,16 @@ def generate_codelist_results(self):
...
-        csvw_state = self.state.csvw_state
+        csvw_inspector = self.state.csvw_inspector
         self.result_code_list_cols = select_codelist_cols_by_csv_url(
-            csvw_state.rdf_graph, self.primary_csv_url
+            csvw_inspector.rdf_graph, self.primary_csv_url
         )
         # Retrieving the primary key column names of the code list to identify the unique identifier
-        result_primary_key_col_names_by_csv_url: PrimaryKeyColNamesByDatasetUrlResult = select_primary_key_col_names_by_csv_url(
-            csvw_state.rdf_graph, self.primary_csv_url
+        # result_primary_key_col_names_by_csv_url: PrimaryKeyColNamesByDatasetUrlResult = select_primary_key_col_names_by_csv_url(
```
robons, line 211: Can we remove this commented out code now?

src/csvcubed/readers/skoscodelistreader.py
```diff
@@ -30,8 +31,10 @@ def extract_code_list_concept_scheme_info(
       `concept_uri_template` uses the standard `notation` uri template variable even if the underlying file uses a
        different column name.
     """
+
     csvw_rdf_manager = CsvwRdfManager(code_list_csvw_path)
-    result = select_table_schema_properties(csvw_rdf_manager.rdf_graph)
+    csvw_inspector = CsvWInspector(csvw_rdf_manager.rdf_graph, code_list_csvw_path)
```
robons, line 36: Instead of creating it yourselves, can you access it via `csvw_rdf_manager.csvw_inspector` so we (hopefully) only ever create one `CsvWInspector` per CSV-W loaded? 

If we have multiple instances of the `CsvWInspector` class, we'll still end up running queries multiple times even with the `@cached_property`.

src/csvcubed/cli/inspect/metadataprinter.py
```diff
@@ -203,13 +203,16 @@ def generate_codelist_results(self):
...
+        #     csvw_inspector.rdf_graph, self.primary_csv_url
+        # )
+        result_primary_key_col_names_by_csv_url: PrimaryKeyColNamesByDatasetUrlResult = (
+            self.state.csvw_inspector.get_codelist_primary_key_by_csv_url()
```
robons, line 215: Can this information (the primary key columns) be tacked on to the `get_table_schema_properties` query?

At any rate, I think the name of it needs to have the worse `codelist` removed since it isn't specifically about code lists (if it were it'd be on the CodeListInspector).

src/csvcubed/utils/sparql_handler/csvw_inspector.py
```diff
@@ -48,6 +54,26 @@ def csvw_type(self) -> CSVWType:
                 "The input metadata is invalid as it is not a data cube or a code list."
             )
 
+    @cached_property
+    def _table_schema_properties(self) -> TableSchemaPropertiesResult:
+        """
+        Cached property for the select_table_schema_properties query that stores the query's results.
+        """
+        result = select_table_schema_properties(self.rdf_graph)
```
robons, line 62: Shouldn't each csv inside the CSV-W have their own table schema properties? i.e. I think this should be a dictionary mapping `csv_url` to `TableSchemaPropertiesResult`.

src/csvcubed/utils/sparql_handler/csvw_inspector.py
```diff
@@ -48,6 +54,26 @@ def csvw_type(self) -> CSVWType:
...
+        return result
+
+    @cached_property
+    def _codelist_primary_key_by_csv_url(self) -> PrimaryKeyColNamesByDatasetUrlResult:
+        """
+        Cached property for the select_primary_key_col_names_by_csv_url query that stores the query's results.
+        """
+        csv_url: str
+        result = select_primary_key_col_names_by_csv_url(
```
robons, line 71: Again, I think we should have this information for each of the csvs present in the CSV-W. I think we should try to combine this query into `_table_schema_properties`.

To be honest, I think we should rename `_table_schema_properties` to `_tables` since it's effectively listing all of the tables in our CSV-W (with the additional information).

src/csvcubed/utils/sparql_handler/csvw_inspector.py
```diff
@@ -59,3 +85,17 @@ def get_primary_catalog_metadata(self) -> CatalogMetadataResult:
         raise KeyError(
             f"Could not find catalog metadata in primary graph '{self.primary_graph_uri}'."
         )
+
+    def get_table_schema_properties(self) -> TableSchemaPropertiesResult:
```
robons, line 89: I think this should be `get_table_info_for_csv_url(self, csv_url: str)....`

tests/unit/utils/sparqlhandler/test_csvw_inspector.py
```diff
@@ -55,5 +56,48 @@ def test_detect_csvw_type_code_list():
...
+    Ensures that the correct table schema properties are returned for the given code list.
+    """
+    csvw_metadata_json_path = (
+        _test_case_base_dir
+        / "pivoted-single-measure-dataset"
+        / "some-dimension.csv-metadata.json"
+    )
+    csvw_rdf_manager = CsvwRdfManager(csvw_metadata_json_path)
+
+    csvw_inspector: CsvWInspector = CsvWInspector(
```
robons, line 74: We should probably agree on some consistent case sensitivity of `CsvW` or `Csvw` in class names to make it more predictable. Any thoughts?

robons (changes requested):

tests/unit/utils/sparqlhandler/test_csvw_inspector.py
SarahJohnsonONS: Personal preference for CsvW

src/csvcubed/cli/inspect/metadataprinter.py
SarahJohnsonONS: Done

src/csvcubed/utils/sparql_handler/csvw_inspector.py
SarahJohnsonONS: Done

src/csvcubed/utils/sparql_handler/csvw_inspector.py
SarahJohnsonONS: Done

src/csvcubed/utils/sparql_handler/csvw_inspector.py
```diff
@@ -69,12 +97,11 @@ def get_primary_catalog_metadata(self) -> CatalogMetadataResult:
             f"Could not find catalog metadata in primary graph '{self.primary_graph_uri}'."
         )
 
-    def get_column_definitions_for_csv(self, csv_url: str) -> List[ColumnDefinition]:
+    def get_table_info_for_csv_url(self, csv_url: str) -> TableSchemaPropertiesResult:
         """
-        Returns the `ColumnDefinition`s for a given csv file, raises a KeyError if the csv_url
```
robons, line 74: Can you go back to the comment you've replaced here.

src/csvcubed/utils/sparql_handler/sparqlquerymanager.py
```diff
@@ -391,23 +392,17 @@ def select_metadata_dependencies(
 
 def select_table_schema_properties(
     rdf_graph: rdflib.Graph,
-) -> TableSchemaPropertiesResult:
+) -> List[TableSchemaPropertiesResult]:
     """
+    TODO: Change description
```
robons, line 397: todo

src/csvcubed/utils/sparql_handler/data_cube_state.py
```diff
@@ -8,38 +8,60 @@
...
+from csvcubed.utils.sparql_handler.csvw_inspector import CsvWInspector
 from csvcubed.utils.sparql_handler.sparqlquerymanager import (
+    select_column_definitions,
     select_csvw_dsd_qube_components,
     select_data_set_dsd_and_csv_url,
     select_is_pivoted_shape_for_measures_in_data_set,
     select_units,
 )
 
+T = TypeVar("T")
```
robons, line 32: What do we need this `T` for?

src/csvcubed/utils/sparql_handler/data_cube_state.py
```diff
@@ -8,38 +8,60 @@
...
 @dataclass
 class DataCubeState:
-    """Provides access to inspect the data cubes contained in an rdflib graph."""
+    csvw_inspector: CsvWInspector
+
+    """
+    Private utility functions.
+    """
+
+    def _get_value_for_key(self, key: str, dict: Dict[str, T]) -> T:
```
robons, line 43: Please don't use this. Use the function we have in `csvcubed.utils.dict` (`get_from_dict_ensure_exists`) instead since it does the same thing.

We deliberately removed this function from this class.

src/csvcubed/utils/sparql_handler/data_cube_state.py
```diff
@@ -8,38 +8,60 @@
...
+        Map of csv_url to the list of column definitions for the given CSV file.
+        """
+        results = select_column_definitions(self.csvw_inspector.rdf_graph)
+        return group_by(results, lambda r: r.csv_url)
 
     @cached_property
     def _units(self) -> Dict[str, UnitResult]:
-        """Gets the unit_uri for each UnitResult"""
-        results = select_units(self.csvw_state.rdf_graph)
+        """ """
```
robons, line 63: Can you bring the comment back please?

src/csvcubed/utils/sparql_handler/sparqlquerymanager.py
```diff
@@ -34,6 +34,7 @@
     PrimaryKeyColNamesByDatasetUrlResult,
     QubeComponentsResult,
     TableSchemaPropertiesResult,
+    TableSchemaPropertiesResults,
```
robons, line 37: Do we  use this class at all?

tests/unit/utils/sparqlhandler/test_csvw_inspector.py
```diff
@@ -55,49 +54,52 @@ def test_get_primary_catalog_metadata_key_error():
...
     )
-    csvw_rdf_manager = get_csvw_rdf_manager(csvw_metadata_json_path)
+    csvw_rdf_manager = CsvwRdfManager(csvw_metadata_json_path)
 
-    csvw_type = csvw_rdf_manager.csvw_state.csvw_type
+    csvw_type = csvw_rdf_manager.csvw_inspector.csvw_type
     assert csvw_type == CSVWType.QbDataSet
 
 
-def test_csvw_type_key_error():
```
robons, line 71: Where has this key error test gone? Do we lose anything without it?

tests/unit/utils/sparqlhandler/test_csvw_inspector.py
robons: We should probably rename the `CsvwRdfManager` to `CsvWRdfManager` then.

tests/unit/utils/sparqlhandler/test_csvw_inspector.py
```diff
@@ -55,49 +54,52 @@ def test_get_primary_catalog_metadata_key_error():
...
-    assert csvw_type == CSVWType.CodeList
+    csvw_inspector: CsvWInspector = CsvWInspector(
+        csvw_rdf_manager.rdf_graph, csvw_metadata_json_path
+    )
+
+    result = csvw_inspector.get_table_info_for_csv_url("some-dimension.csv")
+
+    assert result.about_url == "some-dimension.csv#{+uri_identifier}"
+    assert result.csv_url == "some-dimension.csv"
+    assert result.primary_key_col_names == ["uri_identifier"]
```
robons, line 105: Can you add a test for a CSV-W with multiple columns in the primary key?

If you use a cube (with locally defined code lists) then you'll be able to test a CSV-W that has multiple tables in it as well - which is another test I'd like to see.

src/csvcubed/cli/inspect/metadataprinter.py
```diff
@@ -190,18 +190,20 @@ def generate_codelist_results(self):
...
-        result_primary_key_col_names_by_csv_url: PrimaryKeyColNamesByDatasetUrlResult = select_primary_key_col_names_by_csv_url(
-            csvw_state.rdf_graph, self.primary_csv_url
-        )
-        primary_key_col_names = (
-            result_primary_key_col_names_by_csv_url.primary_key_col_names
+        result_table_schema_properties = csvw_inspector.get_table_info_for_csv_url(
+            self.primary_csv_url
         )
 
+        # result_primary_key_col_names_by_csv_url: PrimaryKeyColNamesByDatasetUrlResult = (
```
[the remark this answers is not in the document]
SarahJohnsonONS: Done

src/csvcubed/models/sparqlresults.py
```diff
@@ -237,8 +237,15 @@ class TableSchemaPropertiesResult:
...
 
     about_url: str
-    value_url: str
-    table_url: str
+    csv_url: str
+    primary_key_col_names: List[str]
+
+
+@dataclass
+class TableSchemaPropertiesResults:
```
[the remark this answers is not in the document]
SarahJohnsonONS: Removed

src/csvcubed/models/sparqlresults.py
```diff
@@ -676,24 +683,19 @@ def map_row(row_result: Dict[str, Any]) -> MetadataDependenciesResult:
...
-    sparql_result: ResultRow,
-) -> TableSchemaPropertiesResult:
-    """
-    Maps sparql query result to `TableSchemaPropertiesResult`
-
-    Member of :file:`./models/sparqlresults.py`
+def map_table_schema_properties_results(
+    sparql_results: List[ResultRow],
+) -> List[TableSchemaPropertiesResult]:
+    """ """
```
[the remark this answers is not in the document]
SarahJohnsonONS: Done

src/csvcubed/utils/sparql_handler/csvw_inspector.py
SarahJohnsonONS: Bit confused about this - I don't think we've changed this comment? It's the same as the one that's currently in csvw_state.py on main

src/csvcubed/utils/sparql_handler/sparqlquerymanager.py
SarahJohnsonONS: Done

src/csvcubed/utils/sparql_handler/data_cube_state.py
SarahJohnsonONS: Removed

src/csvcubed/utils/sparql_handler/data_cube_state.py
SarahJohnsonONS: Removed

src/csvcubed/utils/sparql_handler/data_cube_state.py
SarahJohnsonONS: Done

src/csvcubed/utils/sparql_handler/sparqlquerymanager.py
SarahJohnsonONS: No - removed

tests/unit/utils/sparqlhandler/test_csvw_inspector.py
SarahJohnsonONS: Reinstated

tests/unit/utils/sparqlhandler/test_csvw_inspector.py
SarahJohnsonONS: Multiple columns test added. Do we need to add a "get_all_tables" function before we add the second test here?

robons: I don't think so, you can just access the private property.

tests/unit/utils/sparqlhandler/test_csvw_inspector.py
SarahJohnsonONS: Done

tests/unit/utils/sparqlhandler/test_csvw_inspector.py
SarahJohnsonONS: Done

src/csvcubed/models/sparqlresults.py
```diff
@@ -233,12 +233,12 @@ class MetadataDependenciesResult:
...
     """
-    Model representing the table schema value url and about url.
+    Model representing the about_url, csv_url and a list of primary key column names.
     """
 
     about_url: str
-    value_url: str
-    table_url: str
+    csv_url: str
+    primary_key_col_names: List[str]
```
robons, line 241: Could you either change this to a set or put a docstring comment underneath it explaining that the order is not guaranteed?

src/csvcubed/readers/skoscodelistreader.py
```diff
@@ -12,10 +12,8 @@
 from uritemplate import variables
 
 from csvcubed.utils.iterables import first
-from csvcubed.utils.sparql_handler.sparqlquerymanager import (
-    select_table_schema_properties,
-)
-from csvcubed.utils.tableschema import CsvwRdfManager
+from csvcubed.utils.sparql_handler.code_list_inspector import CodeListInspector
+from tests.helpers.inspectors_cache import get_csvw_rdf_manager
```
robons, line 16: This test code should not be used anywhere outside tests. The cache is *only* for testing.

src/csvcubed/utils/sparql_handler/csvw_inspector.py
robons: The previous comment was:

> Returns the `ColumnDefinition`s for a given csv file, raises a KeyError if the csv_url is not associated with a ColumnDefinition.

And this PR replaces it with:

> Retrieves the stored result of the table schema properties cached property.

src/csvcubed/utils/sparql_handler/data_cube_state.py
```diff
@@ -30,16 +32,26 @@
 
 @dataclass
 class DataCubeState:
-    """Provides access to inspect the data cubes contained in an rdflib graph."""
```
robons, line 33: Can you stick this docstring comment back?

src/csvcubed/utils/sparql_handler/sparqlquerymanager.py
robons: Seems like the `TODO` is still in here.

tests/unit/utils/sparqlhandler/test_csvw_inspector.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+def test_get_table_info_multiple_primary_keys():
+    """
+    Ensures that primary keys can be retrieved successfully from a table that contains multiple primary keys.
```
robons, line 129: This shouldn't specifically be about primary keys.

And could you please remove the bit of the comment which talks about the ordering of the primary keys since we realised it wasn't possible to extract them with the order maintained from the RDF graph.

robons (changes requested):
Almost there. The main problem I have is with the use of the testing code list inspector cache function being used in production code.

N.B. Code in the `tests` package doesn't end up making it to the user's machine, so this functionality would break on an end-user's machine.

tests/unit/utils/sparqlhandler/test_csvw_inspector.py
NickPapONS: We decided to remove this unit test because we realized the new test (below it) now covers everything this test did, but also asserts a few extra things, so we think this test is now redundant. Though we can put it back and change it if that is best.

src/csvcubed/utils/sparql_handler/csvw_inspector.py
SarahJohnsonONS: "Retrieves the stored result of the table schema properties cached property" is the comment for `get_table_info_for_csv_url` not `get_column_definitions_for_csv`

robons: Ah ok, my bad

robons (approved):

PR #734, opened 2023-02-20
Ridding us of the bad refPeriod code lists

canwaf:
We need a better solution than this, until then we won't generate useless code lists.


robons:
I was under the impression that we would automatically generate code lists for date-time periods, and that by setting codelist to false we're stopping people from being able to filter by the datetime dimensions?

I believe this was a much requested bit of functionality. Why do you want to turn it off?

Just to remind you that you can still override how templates work in individual data sets.

Issue #739, opened 2023-02-21
[BUG] Themes duplicated 1 time for every keyword when using inspect command


canwaf:
**Describe the issue:**
This one is a weird one, but here goes:

It appears that the theme in a data set's inspect will be duplicated in the inspect output for each keyword in a data set. The CSV-W does serialise properly though. Example from CSV-W built using input files at https://github.com/GSS-Cogs/provider-ons-ess/tree/459cb3b69c548fc45ca3545cb34adf4690ed72f4/aged-16-to-64-years-level-3-or-above-qualifications. Full output at https://gss-cogs.github.io/provider-ons-ess/out/aged-16-to-64-years-level-3-or-above-qualifications/aged-16-to-64-years-level-3-or-above-qualifications/aged-16-to-64-years-level-3-or-above-qualifications.csv-metadata.json_inspect_output.txt.

```
        - Themes: 
                -- https://www.ons.gov.uk/employmentandlabourmarket/peopleinwork
                -- https://www.ons.gov.uk/employmentandlabourmarket/peopleinwork
                -- https://www.ons.gov.uk/employmentandlabourmarket/peopleinwork
                -- https://www.ons.gov.uk/employmentandlabourmarket/peopleinwork
                -- https://www.ons.gov.uk/employmentandlabourmarket/peopleinwork
                -- https://www.ons.gov.uk/employmentandlabourmarket/peopleinwork
                -- https://www.ons.gov.uk/employmentandlabourmarket/peopleinwork
        - Keywords: 
                -- levelling-up
                -- subnational
                -- education
                -- qualifications
                -- united-kingdom
                -- mission-6
                -- employment
```

**Reproducible steps / Code example:**
`csvcubed inspect out/aged-16-to-64-years-level-3-or-above-qualifications.csv-metadata.json`

**Error message obtained / Resulting behaviour:**
As above.

**Expected behaviour:**
A 1:1 relationship between themes and themes. Not a 1:1 relationship between themes and keywords.

**Additional Information:**
This one makes me happy because I haven't a single clue HOW this happens. But I experimented by adding new keywords and the repeat of themes just increased.

**Version information + Operating System:**
python 3.10.3, csvcubed 0.2.3, MacOS something or other.

robons:
Sounds like a cross-join in a SPARQL query bringing out more than we'd hoped for.

SarahJohnsonONS:
Attempt at bug triage:

Impact - 2 or 3
[...]

PR #746, opened 2023-03-01
#739 Inspect API - Fixing duplicated keywords + themes in CatalogMetadata

nimshi89:
Changed the select_catalog_metadata.sparql to return distinct values of themes, keyword and landing pages.

Satisfies ticket #739.


NickPapONS (commented):
Looks good! The change to the sparql query makes sense and I like seeing the whole team's work on the cached properties come into effect by making that unit test simpler/more efficient.

tests/unit/utils/sparqlhandler/test_csvw_inspector.py
```diff
@@ -256,3 +256,34 @@ def test_get_column_definitions_for_csv():
         value_url="http://example.com/value",
         virtual=True,
     )
+
+
+def test_multi_theme_and_keyword():
+    """
+    This test ensures that the 'select_catalog_metadata.sparql' will return the correct amount of
+    themes, keywords and landing pages(unique values and no duplications)
```
robons, line 264: It'd be helpful if you could stick the URL to the bug in here.

robons (approved):
Looks good.

It might be helpful to alter the test case you've got to artificially add more themes (and some landing pages).

tests/unit/utils/sparqlhandler/test_csvw_inspector.py
nimshi89: yeah sure

Issue #674, opened 2022-12-20
Implement validations in the following classes


nimshi89:

robons:
@MuazzamChaud this might be a goood ticket to start work on whilst I'm away. Speak with @nimshi89 if you're uncertain about anything or need any support.

PR #751, opened 2023-03-03
#674 Finishing responding to feedback for validations pt1

NickPapONS:
Finishing up responding to feedback for #674 Implement validations in the following classes PT1.


robons (approved):
Thanks for tidying that up!

Issue #675, opened 2022-12-20
Implement validations in the following classes PT2.


nimshi89:
Implement the validations on the following classes as shown in the Dimensions class. Any new validations will require the appropriate test to be implemented as well.

https://github.com/GSS-Cogs/csvcubed/blob/5f2989f69c03c22b99bec656f86d2303af0c258e/src/csvcubed/models/cube/qb/components/dimension.py#L43

```
Cube(Generic[TMetadata], PydanticModel) (csvcubed.models.cube.cube)
CsvColumn(PydanticModel, UriIdentifiable, ABC) (csvcubed.models.cube.columns)
    SuppressedCsvColumn(CsvColumn) (csvcubed.models.cube.columns)
    QbColumn(CsvColumn, Generic[QbColumnStructuralDefinition]) (csvcubed.models.cube.qb.columns)
CatalogMetadataBase(PydanticModel, ABC) (csvcubed.models.cube.catalog)
    CatalogMetadata(CatalogMetadataBase, UriIdentifiable) (csvcubed.models.cube.qb.catalog)
TripleFragmentBase(PydanticModel, ABC) (csvcubed.models.cube.qb.components.arbitraryrdf)
    TripleFragment(TripleFragmentBase) (csvcubed.models.cube.qb.components.arbitraryrdf)
    InverseTripleFragment(TripleFragmentBase) (csvcubed.models.cube.qb.components.arbitraryrdf)
QbStructuralDefinition(PydanticModel, ABC) (csvcubed.models.cube.qb.components.datastructuredefinition)
    QbColumnStructuralDefinition(QbStructuralDefinition, ABC) (csvcubed.models.cube.qb.components.datastructuredefinition)
        QbAttribute(QbColumnStructuralDefinition, ArbitraryRdf, ABC) (csvcubed.models.cube.qb.components.attribute)
            ExistingQbAttribute(QbAttribute) (csvcubed.models.cube.qb.components.attribute)
                ExistingQbAttributeLiteral(ExistingQbAttribute, QbAttributeLiteral) (csvcubed.models.cube.qb.components.attribute)
            NewQbAttribute(QbAttribute, UriIdentifiable) (csvcubed.models.cube.qb.components.attribute)
                NewQbAttributeLiteral(NewQbAttribute, QbAttributeLiteral) (csvcubed.models.cube.qb.components.attribute)
            QbAttributeLiteral(QbAttribute, ABC) (csvcubed.models.cube.qb.components.attribute)
[...]

PR #752, opened 2023-03-06
#675 Validations implementation pt2.

nimshi89:
This ticket we added several new validations to the second half of classes that were identified as needing new ones. The PR is a follow up to the #674 "Implement validations in the following classes PT1."

Satisfies #675


robons (approved):
Looks good to me!

Issue #755, opened 2023-03-08
There are multiple units components in the DSD of pivoted multi-measure cubes


nimshi89:
This ticket is related to #703. There has been an error where the unit's component is duplicated in pivoted multi-measure cubes. This causes the problem that we cannot uniquely (and correctly) link the units component to the relevant column.

In dsdtordfmodelhelper.py  the modification has been made (by Rob) to avoid the unit column duplication.

```python
   def _get_qb_units_column_specification(
        self, column_name_uri_safe: str
    ) -> List[rdf.qb.AttributeComponentSpecification]:
        if self._units_component_already_defined:
            # Don't define a second units component, the first one will work just fine.
            return []

        self._units_component_already_defined = True
```

This method ensures that if a units column has already been defined, then another column cannot be recorded.

Issue #703, opened 2023-01-23
Add way to map between CSV columns and the components of the cube


CharlesRendle:
**In refinement we may (definitely) want to change the ticket title to something more descriptive**

Further to the work from ticket #690, we want to offer the user the ability to return information about their dataset. This comes in two forms:

1. The users provides the csv_url and gets a list/map of the column name to the component type (+ some column definition information in some fashion).
2. The user presents a particular component type and the API function returns all the `ColumnDefinition` **objects**  of that particular component type.

In the second case, we want to return extra column information instead of just the column titles which is why we should return column objects. E.g. returning attribute column titles in a pivoted shape cube is meaningless without also telling the user which obs val column the attribute is wedded to.

## Implementation Ideas

We *should* be able to make use of the information on the `QubeComponentResult` models which as part of the `DataCubeInspector._dsd_qube_components(...)` function:
   * This identifies which CSV col is a measure column, which is a units columns, which is a dimension and which is an attribute (general) column but does not directly specify which columns are obs val columns.
   * We can figure out which columns are obs val columns by identifying the csv columns which are *not* dimension/measure/units/attributes columns (i.e. the set difference between all columns and the ones with types already identifies). If this doesn't prove to be a sensible approach we can likely derive this information from the `ColumnDefinition` objects that we have available on the `CsvwInspector` class.

## Required Work

- [x] Implement a new API function on the `DataCubeInspector` class which makes use of the existing queries (discussed above) which takes the `csv_url` the user is interested in an returns a list of `ColumnComponentInfo(ComponentType.Attribute, ColumnDefinition(...))` objects -- this list must be returned in the same order in which columns are defined inside the CSV-W/CSV file (the fact we're using a list over a map if because we want to preserve the implicit ordering information).
- [x] Add a second API function to the `DataCubeInspector`, e.g. `get_columns_for_component_type(ComponentType.Dimension)` which returns the `ColumnDefinitions` *in the same order in which they're defined in the CSV/CSV-W* which match the given component type. N.B. I don't think you should return a measure in the list of dimensions, and I don't think you should return a units column in the list of attributes (debatable?).
- [x] Add unit tests to ensure both API functions behave as expected for at minimum the following cube shapes (you should ensure you don't need to test any more than this yourself when doing the ticket): Standard shape cube, Old-style single measure pivoted & New-style multi-measure pivoted data set. You should make sure the ordering requirements are tested as well. 
[...]

PR #756, opened 2023-03-08
#755 FIX unit component duplication BUG

nimshi89:
#755 Bug was noticed during the work on #703.

### DIFF
--- a/src/csvcubed/writers/helpers/qbwriter/dsdtordfmodelshelper.py
+++ b/src/csvcubed/writers/helpers/qbwriter/dsdtordfmodelshelper.py
@@ -74,6 +74,10 @@ class DsdToRdfModelsHelper:
     cube: QbCube
     _uris: UriHelper
     _units_component_already_defined: bool = field(init=False, default=False)
+    """
+    Records whether or not a units component has already been defined in this cube.
+    If it has, don't define it again.
+    """
 
     def generate_data_structure_definitions(self) -> List[dict]:
         """
@@ -310,17 +314,12 @@ class DsdToRdfModelsHelper:
         else:
             raise TypeError(f"Unhandled component type {type(component)}")
 
-    """
-    Records whether or not a units component has already been defined in this cube.
-    If it has, don't define it again.
-    """
-
     def _get_qb_units_column_specification(
         self, column_name_uri_safe: str
     ) -> List[rdf.qb.AttributeComponentSpecification]:
         if self._units_component_already_defined:
             _logger.debug(
-                "The component type has already been serialised, %s.",
+                "Units component already generated. Not generating a second one, %s.",
                 column_name_uri_safe,
             )
             # Don't define a second units component, the first one will work just fine.
--- a/src/csvcubed/writers/helpers/qbwriter/dsdtordfmodelshelper.py
+++ b/src/csvcubed/writers/helpers/qbwriter/dsdtordfmodelshelper.py
@@ -5,7 +5,7 @@ QB DSD Helper
 Help Generate the DSD necessary for an RDF Data Cube.
 """
 import logging
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Iterable, List, Set
 
 from csvcubedmodels import rdf
@@ -73,6 +73,7 @@ _logger = logging.getLogger(__name__)
 class DsdToRdfModelsHelper:
     cube: QbCube
     _uris: UriHelper
+    _units_component_already_defined: bool = field(init=False, default=False)
 
     def generate_data_structure_definitions(self) -> List[dict]:
         """
@@ -309,7 +310,6 @@ class DsdToRdfModelsHelper:
         else:
             raise TypeError(f"Unhandled component type {type(component)}")
 
-    _units_component_already_defined: bool = False
     """
     Records whether or not a units component has already been defined in this cube.
     If it has, don't define it again.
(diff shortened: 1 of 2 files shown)


NickPapONS (commented):
Looks good to me! All I would do is tell sonarcloud that the "Existing Dimension Component" used in the new unit test is not an issue so the code smell disappears.

src/csvcubed/writers/helpers/qbwriter/dsdtordfmodelshelper.py
```diff
@@ -301,17 +301,29 @@ def _get_qb_component_specs_for_col(
...
-            return [self._get_qb_units_column_specification(column_name_uri_safe)]
+            return self._get_qb_units_column_specification(column_name_uri_safe)
         elif isinstance(component, QbMultiMeasureDimension):
             return self._get_qb_measure_dimension_specifications(component)
         elif isinstance(component, QbObservationValue):
             return self._get_qb_obs_val_specifications(component)
         else:
             raise TypeError(f"Unhandled component type {type(component)}")
 
+    _units_component_already_defined: bool = False
```
robons, line 312: Could you move this up to the top of the class where the other attributes are defined. Can you also declare that this isn't an attribute which should appear in the `__init__` function? The dev using this class shouldn't be allowed to configure this themselves.

src/csvcubed/writers/helpers/qbwriter/dsdtordfmodelshelper.py
```diff
@@ -301,17 +301,29 @@ def _get_qb_component_specs_for_col(
...
+    If it has, don't define it again.
+    """
+
     def _get_qb_units_column_specification(
         self, column_name_uri_safe: str
-    ) -> rdf.qb.AttributeComponentSpecification:
+    ) -> List[rdf.qb.AttributeComponentSpecification]:
+        if self._units_component_already_defined:
+            # Don't define a second units component, the first one will work just fine.
+            return []
```
robons, line 323: Can you add some debug logging here so we can track that we're avoiding serialising a second units component?

tests/unit/writers/qbwriter/test_dsdtordfmodelshelper.py
```diff
@@ -577,5 +578,69 @@ def test_qb_order_of_components():
...
+                QbObservationValue(
+                    NewQbMeasure("Some Other Measure"), NewQbUnit("Some Other Unit")
+                ),
+            ),
+        ],
+    )
+
+    dsd_helper = DsdToRdfModelsHelper(cube, UriHelper(cube))
+    dataset = dsd_helper._generate_qb_dataset_dsd_definitions()
+    graph = dataset.to_graph(Graph())
```
robons, line 624: We don't need this graph or the assertion do we?

src/csvcubed/writers/helpers/qbwriter/dsdtordfmodelshelper.py
nimshi89: sorted

src/csvcubed/writers/helpers/qbwriter/dsdtordfmodelshelper.py
nimshi89:
```    def _get_qb_units_column_specification(
        self, column_name_uri_safe: str
    ) -> List[rdf.qb.AttributeComponentSpecification]:
        if self._units_component_already_defined:
            _logger.debug(
                "The component type has already been serialised, %s.",
                column_name_uri_safe,
            )
            # Don't define a second units component, the first one will work just fine.

            return []```

is this the right place for it?

File: tests/unit/writers/qbwriter/test_dsdtordfmodelshelper.py (hunk 3 above)
Comment (nimshi89) [reply tests/unit/writers/qbwriter/test_dsdtordfmodelshelper.py]: sorted

File: src/csvcubed/writers/helpers/qbwriter/dsdtordfmodelshelper.py [hunk 4]
```diff
@@ -301,17 +302,33 @@ def _get_qb_component_specs_for_col(
...
-            return [self._get_qb_units_column_specification(column_name_uri_safe)]
+            return self._get_qb_units_column_specification(column_name_uri_safe)
         elif isinstance(component, QbMultiMeasureDimension):
             return self._get_qb_measure_dimension_specifications(component)
         elif isinstance(component, QbObservationValue):
             return self._get_qb_obs_val_specifications(component)
         else:
             raise TypeError(f"Unhandled component type {type(component)}")
 
+    """
```
robons, line 313: This comment should be against the field we added (which you've now moved up).

src/csvcubed/writers/helpers/qbwriter/dsdtordfmodelshelper.py
robons: that's definitely the right place, but could it say something like `Units component already generated. Not generating a second one.`

nimshi89: Yeah cool I'll sort it out

src/csvcubed/writers/helpers/qbwriter/dsdtordfmodelshelper.py
nimshi89: moved it

robons (approved):

Issue #520, opened 2022-06-09
Add "How to write metadata" guide or procedure


canwaf:
Add a new page of documentation guiding a user through how to write their metadata in the qube-config.json file. The current file is a bit too technical for most users and doesn't provide them with examples.

- [ ] Guide users through all of the fields they can use, explain to them what the field contains and provide examples for them to choose from.
- [ ] Link back to the existing qube-config.json for more advanced configuration options.
- [ ] Explain what the squiggly lines in vscode mean and show auto-completion
- [ ] Give them lots of examples of JSON documents they can copy & paste.

- Probably want multiple pages
- Guide to defining the metadata
- Guide to defining columns

See the below from #706 for more thoughts and ideas. This seems to be a duplicate ticket.

------------

Following on from the completion of #648, the following suggestions have been identified to improve the user guidance on writing a qube-config.json file:

[General comments:](https://gss-cogs.github.io/csvcubed-docs/external/guides/configuration/qube-config/)

- [ ] Resolve whether the terminology is “qube-config” or “cube-config” (all pages) (include in [Confluence Style Guide](https://confluence.ons.gov.uk/display/PHEM/Style+Guide))
- [ ] Resolve whether csv and uri should be capitalised or not (all pages) (include in [Confluence Style Guide](https://confluence.ons.gov.uk/display/PHEM/Style+Guide))
- [ ] Include a link to the current qube-config schema
- [ ] Use one data set throughout guide (if feasible)

[Column definitions](https://gss-cogs.github.io/csvcubed-docs/external/guides/configuration/qube-config/#column-definitions)

- [ ] “Using existing columns” section could do with better explanations - e.g. “Unless the component being reused is a literal attribute and you're providing a "data_type" key-value pair, any other key-value pairs provided will change the column to a new component which will extend the linked parent component.”
- [ ] For components other than Dimension - make it clearer that “type” will default to “dimension” unless explicitly stated
- [ ] Include explanations about New/Existing - existing where?
- [ ] Reorder tables to match the order in the schema

[Shared column configuration option](https://gss-cogs.github.io/csvcubed-docs/external/guides/configuration/qube-config/#shared-column-configuration-options)

- [ ] Repetitive - could be incorporated into individual sections

[Measures and Units](https://gss-cogs.github.io/csvcubed-docs/external/guides/configuration/qube-config/#measures-and-units)

- [ ] Make it clearer that the properties in the table go inside the “values” property for measures/units

[...]

PR #766, opened 2023-03-28
New attribute literal `label` set as Optional

SarahJohnsonONS:
Bug fix - changed NewAttributeLiteral `label` to Optional[str] - if `label` not provided, it defaults to the column title.

### DIFF
--- a/src/csvcubed/readers/cubeconfig/v1/columnschema.py
+++ b/src/csvcubed/readers/cubeconfig/v1/columnschema.py
@@ -278,8 +278,8 @@ class ExistingAttributeResource(SchemaBaseClass):
 
 @dataclass
 class NewAttributeLiteral(SchemaBaseClass):
-    label: str
     data_type: str
+    label: Optional[str] = None
     description: Optional[str] = None
     from_existing: Optional[str] = None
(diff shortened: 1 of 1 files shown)


CharlesRendle (commented):
Happy that this change satisfies the ticket.
Have verified that the column name is used for label property when building a csvw
And it looks like the documentation portion has been addressed in ticket #520
Approving, but worth thinking if this needs a unit test to demonstrate what the problem is/was and the functionality now expected under the bug conditions. E.g. 

When an attribute column is defined:
```
"columns": {
    "My New Attribute Literal": {
         "type": "Attribute",
         "data_type": "decimal",
    }
}
```

The metadata.json file should contain:
```
 "http://www.w3.org/2000/01/rdf-schema#label": [
                {
                    "@language": "en",
                    "@value": "Attribute"
                }
            ],
```

@robons given it's such a small change, what do you think?

CharlesRendle (changes requested):
Spoken with Rob - please can you add a unit test to demonstrate the issue/fix for this ticket.

CharlesRendle (approved):
All looks good to me!
Thank you for writing the test

Issue #747, opened 2023-03-02
Write initial pandas DataFrame Function


robons:
## Functionality to Implement

- [x] Work with the pandas CSV load function in `csvcubed.utils.pandas`(?). Using the information generated by the work in #703, we need to tell pandas what the data type of each column is (e.g. If it's a dimension then it should be a string, if it's an attribute literal or an observation then we need to use the datatype specified in the CSV-W, etc.).
- [x] Write a new function in DataCubeInspector to load the data into a dataframe - this should take the `csv_url` of the cube the user wants to be able to load. If the csv_url doesn't represent a cube then we're free to raise an Exception.

## Testing

For the first test-case, use a CSV-W with:
   - Multiple dimensions
   - An attribute resource column
   - An attribute literal column (with a specified non-default datatype)
   - A units column
   - A measures columns
   - A single obs val column with a specified non-default datatype (standard shape)
 
For the second test-case, use a CSV-W with:
   - Multiple Dimensions
   - Multiple obs val columns (pivoted shape)

- [x] Put both of these CSV-W inputs into the function to load the raw data and ensure that each of the columns has the expected data type and the correct values contained within.

PR #770, opened 2023-03-31
747 pandas df function

CharlesRendle:
PR addressing issue #747

Adds an inspect API function to load a pandas dataframe from the provided `csv_url`
The dataframe loads each column with the correct data type corresponding to the 'column_definition` for obs columns and attribute literal columns. Other columns are assigned as strings.

Two unit tests added to demonstrate this functionality (two test cases also added).

Also adds `get_primary_csv_url()` functions to both code_list_inspector and data_cube_inspector classes which retrieves the csv_url for the primary CSV defined in the CSV-W. Included two unit tests to demonstrate this function (one for each class)

### DIFF
--- a/src/csvcubed/utils/sparql_handler/data_cube_inspector.py
+++ b/src/csvcubed/utils/sparql_handler/data_cube_inspector.py
@@ -8,7 +8,9 @@ one of more data cubes.
 
 from dataclasses import dataclass
 from functools import cache, cached_property
+from pathlib import Path
 from typing import Dict, List, Optional, Set, Tuple
+from urllib.parse import urljoin
 
 import pandas as pd
 
@@ -24,6 +26,7 @@ from csvcubed.models.sparqlresults import (
     QubeComponentsResult,
     UnitResult,
 )
+from csvcubed.models.validationerror import ValidationError
 from csvcubed.utils.dict import get_from_dict_ensure_exists
 from csvcubed.utils.iterables import first, group_by
 from csvcubed.utils.pandas import read_csv
@@ -150,21 +153,6 @@ class DataCubeInspector:
             self.csvw_inspector.csvw_json_path,
         )
 
-    @cached_property
-    def _load_dataframe(self) -> pd.DataFrame:
-        """
-        TODO: This
-        TODO: Do we even want this as a cached property?
-        """
-        # df = read_csv(self.csvw_inspector.get_primary_catalog_metadata().dataset_uri)
-        # df = read_csv(
-        #     self.get_cube_identifiers_for_csv(
-        #         self.csvw_inspector.get_primary_catalog_metadata().dataset_uri
-        #     ).csv_url
-        # )
-        # csv_url = csv_url
-        # return read_csv(csv_url)
-
     """
     Public getters for the cached properties.
     """
@@ -229,13 +217,40 @@ class DataCubeInspector:
         """
         TODO: This and maybe change API function name.
         """
-        # return self._load_dataframe(csv_url)
-        df: pd.DataFrame = read_csv(csv_url)
+        from csvcubed.models.cube.qb.components.constants import (
+            ACCEPTED_DATATYPE_MAPPING,
+        )
 
-        #
+        cols = self.get_column_component_info(csv_url)
+        dict_of_types = {}
+        for col in cols:
+            if (
+                col.column_type.value == "Observations"
+                or col.column_type.value == "Attributes"
+            ):
+                col_data_type = col.column_definition.data_type.removeprefix(
+                    "http://www.w3.org/2001/XMLSchema#"
+                )
+                if col_data_type in ACCEPTED_DATATYPE_MAPPING.keys():
+                    dict_of_types[
+                        col.column_definition.title
+                    ] = ACCEPTED_DATATYPE_MAPPING[col_data_type]
+                else:
+                    # raise an error?
+                    pass
+            else:
+                dict_of_types[col.column_definition.title] = ACCEPTED_DATATYPE_MAPPING[
+                    "string"
+                ]
 
-        # df.astype({}).dtypes
-        return df
+        absolute_csv_url = Path(
+            urljoin(self.csvw_inspector.csvw_json_path.as_uri(), csv_url)
+            .removeprefix("file:\\")
+            .removeprefix("file:")
+        )
+        df: Tuple[pd.DataFrame, List[ValidationError]] = read_csv(absolute_csv_url)
+        df[0].astype(dict_of_types, copy=True).dtypes
+        return df[0]
 
     @cache
     def get_column_component_info(self, csv_url: str) -> List[ColumnComponentInfo]:
@@ -288,6 +303,15 @@ class DataCubeInspector:
             for c in self.get_column_component_info(csv_url)
             if c.column_type == column_type
         ]
+
+    def get_primary_csv_url(self) -> str:
+        """
+        Retrieves the csv_url for the primary CSV defined in the CSV-W.
+        """
+        primary_catalog_metadata = self.csvw_inspector.get_primary_catalog_metadata()
+        return self.get_cube_identifiers_for_data_set(
+            primary_catalog_metadata.dataset_uri
+        ).csv_url
 
 
 def _get_column_type_and_component(
--- a/src/csvcubed/utils/sparql_handler/data_cube_inspector.py
+++ b/src/csvcubed/utils/sparql_handler/data_cube_inspector.py
@@ -260,7 +260,7 @@ class DataCubeInspector:
                 .removeprefix("file:")
             )
         )
-        return read_csv(absolute_csv_url, dtype=dict_of_types)
+        return read_csv(Path(absolute_csv_url), dtype=dict_of_types)
 
     @cache
     def get_column_component_info(self, csv_url: str) -> List[ColumnComponentInfo]:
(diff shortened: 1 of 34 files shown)


src/csvcubed/utils/sparql_handler/data_cube_inspector.py
```diff
@@ -207,6 +217,50 @@ def get_code_lists_and_cols(self, csv_url: str) -> CodelistsResult:
 
         return self._codelists_and_cols.get(csv_url, CodelistsResult([], 0))
 
+    def get_dataframe(self, csv_url: str) -> Tuple[pd.DataFrame, List[ValidationError]]:
+        """
+        Get the pandas dataframe for the csv url of the cube wishing to be loaded.
```
robons, line 222: It's be good to explain what the list of validation errors we return are about in this comment too.

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
```diff
@@ -207,6 +217,50 @@ def get_code_lists_and_cols(self, csv_url: str) -> CodelistsResult:
...
+                        col.column_definition.title
+                    ] = ACCEPTED_DATATYPE_MAPPING[col_data_type]
+                else:
+                    raise ValueError(
+                        f"Unhandled data type '{col.column_definition.data_type}'."
+                    )
+            else:
+                dict_of_types[col.column_definition.title] = "string"
+
+        absolute_csv_url = Path(
```
robons, line 255: Could you refactor this into a utils function which takes a URI and returns a file path? Ideally in the comment you could explain that the function works on both windows and lnux/unix.

There are a couple of other locations in the code base which do this same kind of thing. Could you tidy them up to use this new function too?

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
```diff
@@ -207,6 +217,50 @@ def get_code_lists_and_cols(self, csv_url: str) -> CodelistsResult:
 
         return self._codelists_and_cols.get(csv_url, CodelistsResult([], 0))
 
+    def get_dataframe(self, csv_url: str) -> Tuple[pd.DataFrame, List[ValidationError]]:
+        """
+        Get the pandas dataframe for the csv url of the cube wishing to be loaded.
+        """
+        cols = self.get_column_component_info(csv_url)
+        dict_of_types = {}
+        for col in cols:
```
robons, line 226: This function seems to be getting a bit big now. Could you simplify it a bit by refactoring some of the code into a separate function? Perhaps the stuff which generates the `dict_of_types` makes sense to put into a separate function?

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
```diff
@@ -207,6 +217,50 @@ def get_code_lists_and_cols(self, csv_url: str) -> CodelistsResult:
...
+                dict_of_types[col.column_definition.title] = "string"
+
+        absolute_csv_url = Path(
+            os.path.normpath(
+                urljoin(self.csvw_inspector.csvw_json_path.as_uri(), csv_url)
+            )
+            .removeprefix("file:\\")
+            .removeprefix("file:")
+        )
+        return read_csv(Path(absolute_csv_url), dtype=dict_of_types)
```
robons, line 262: `absolute_csv_url` is already a Path at this point.

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
```diff
@@ -207,6 +217,50 @@ def get_code_lists_and_cols(self, csv_url: str) -> CodelistsResult:
...
+                col.column_type == EndUserColumnType.Observations
+                or is_attribute_literal
+            ):
+                if col.column_definition.data_type is not None:
+                    col_data_type = col.column_definition.data_type.removeprefix(
+                        _XSD_BASE_URI
+                    )
+                else:
+                    raise ValueError(
+                        "Expected a defined datatype but got 'None' instead."
```
robons, line 242: Can you tell us which column it is that's the problem in this message? It'll help debugging when this inevitably happens.

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
```diff
@@ -207,6 +217,50 @@ def get_code_lists_and_cols(self, csv_url: str) -> CodelistsResult:
...
+            is_attribute_literal = (
+                col.column_type == EndUserColumnType.Attribute
+                and col.column_definition.value_url is None
+            )
+
+            if (
+                col.column_type == EndUserColumnType.Observations
+                or is_attribute_literal
+            ):
+                if col.column_definition.data_type is not None:
```
robons, line 236: I think this code would read better if it were like this:

```python
if col.column_definition.data_type is None:
   raise ValueError(...)

col_data_type = col.column_definition.data_type.removeprefix(...)
```

That way the assignment of col_data_type isn't hidden away in an if-block.

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
```diff
@@ -207,6 +217,50 @@ def get_code_lists_and_cols(self, csv_url: str) -> CodelistsResult:
...
+                    col_data_type = col.column_definition.data_type.removeprefix(
+                        _XSD_BASE_URI
+                    )
+                else:
+                    raise ValueError(
+                        "Expected a defined datatype but got 'None' instead."
+                    )
+                if col_data_type in ACCEPTED_DATATYPE_MAPPING.keys():
```
robons, line 244: You can just say `col_data_type in ACCEPTED_DATATYPE_MAPPING`, without the `.keys()` bit here.

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
```diff
@@ -207,6 +217,50 @@ def get_code_lists_and_cols(self, csv_url: str) -> CodelistsResult:
...
+                    dict_of_types[
+                        col.column_definition.title
+                    ] = ACCEPTED_DATATYPE_MAPPING[col_data_type]
+                else:
+                    raise ValueError(
+                        f"Unhandled data type '{col.column_definition.data_type}'."
```
robons, line 250: Again, can you mention the column title in this error message?

tests/unit/utils/sparqlhandler/test_data_cube_inspector.py
```diff
@@ -902,3 +906,73 @@ def test_get_columns_for_component_attribute_pivoted():
...
+        / "testing-converting-a-pivoted-csvw-to-pandas-dataframe.csv-metadata.json"
+    )
+    data_cube_inspector = get_data_cube_inspector(csvw_metadata_json_path)
+    csv_url = data_cube_inspector.get_primary_csv_url()
+
+    dataframe, validation_errors = data_cube_inspector.get_dataframe(csv_url)
+
+    assert isinstance(dataframe, pd.DataFrame)
+    assert dataframe["Dim1"].dtype == "string"
+    assert dataframe["Obs1"].dtype == "string"
```
robons, line 975: Is this the one that has a obs val column which has the time data type?

robons (commented):
Looks pretty close to me, just a bit of refactoring if you could.

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
CharlesRendle: Added further comment.

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
CharlesRendle: refactored the `dict_of_types` bit into a private function.

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
CharlesRendle: removed

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
CharlesRendle: added the column title to the message

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
CharlesRendle: rearranged logic blocks

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
CharlesRendle: removed `.keys()`

src/csvcubed/utils/sparql_handler/data_cube_inspector.py
CharlesRendle: added the column title to the message

tests/unit/utils/sparqlhandler/test_data_cube_inspector.py
CharlesRendle: Yeah it was the `time` column. Have added a comment to explain why we expect "string"

tests/unit/utils/test_uri.py
```diff
@@ -53,5 +57,24 @@ def test_ensure_all_look_like_uri():
...
+def test_get_absolute_file_path():
+    """
+    Testing the URI util function `get_absolute_file_path()` returns a normalised
+    path from a string URI which can be used to locate resources on windows as
+    well as unix/linux operating systems.
+    """
+    csv_url = "file:///workspaces/csvcubed/tests/test-cases/cli/inspect/inspector-load-dataframe/pivoted-shape/pivoted-shape-out/testing-converting-a-pivoted-csvw-to-pandas-dataframe.csv"
+    absolute_csv_url = get_absolute_file_path(csv_url)
+    assert isinstance(absolute_csv_url, Path)
+    if os.name == "nt":
```
robons, line 69: I think you should have different `csv_urls` for each OS. The windows one should probably start with`file:/C:/...`.

robons (approved):
Approved, but could you address the two small comments I've made?

Issue #769, opened 2023-03-31
[BUG] 'MetadataPrinter' object has no attribute 'primary_csv_column_definitions' when inspecting a code list


CharlesRendle:
**Describe the issue:**
The [csv-to-csvw action encountered an error](https://github.com/GSS-Cogs/family-ess/actions/runs/4574366932/jobs/8075936329) when trying to inspect a code list csvw created by csvcubed build. 

**Reproducible steps / Code example:**
Running the inspect command on any code list csvw seems to return the same error. In our test cases folder I've used `cli/inspect/pivoted-single-measure-dataset/some-dimension.csv-metadata.json`

`csvcubed inspect tests/test-cases/cli/inspect/pivoted-single-measure-dataset/some-dimension.csv-metadata.json --log-level debug`

**Error message obtained / Resulting behaviour:**
```
2023-03-31 14:10:35,328 - csvcubed.cli.inspect.inspect - DEBUG - Metadata json-ld path: /workspaces/csvcubed/tests/test-cases/cli/inspect/pivoted-single-measure-dataset/some-dimension.csv-metadata.json (inspect.py:33)
2023-03-31 14:10:35,920 - csvcubed.utils.tableschema - INFO - Successfully parsed csvw json-ld to rdf graph. (tableschema.py:142)
2023-03-31 14:10:35,921 - csvcubed.utils.sparql_handler.sparqlquerymanager - DEBUG - Root path: /workspaces/csvcubed/src/csvcubed (sparqlquerymanager.py:100)
2023-03-31 14:10:35,921 - csvcubed.utils.sparql_handler.sparqlquerymanager - DEBUG - select_csvw_table_schema_file_dependencies query file path: /workspaces/csvcubed/src/csvcubed/utils/sparql_handler/sparql_queries/select_csvw_table_schema_file_dependencies.sparql (sparqlquerymanager.py:109)
2023-03-31 14:10:36,118 - csvcubed.utils.tableschema - DEBUG - Loading dependent file containing table schema some-dimension.table.json into RDF graph. (tableschema.py:70)
2023-03-31 14:10:36,119 - csvcubed.utils.json - DEBUG - Loading JSON from file /workspaces/csvcubed/tests/test-cases/cli/inspect/pivoted-single-measure-dataset/some-dimension.table.json (json.py:62)
[...]

PR #771, opened 2023-04-03
#769 Fixing bug when inspecting code-lists

robons:
Fixing issue introduced in previous release where code-list inspect would fail.

Satisfies issue #769.


CharlesRendle (approved):
Happy this change fixes the issue. Glad to see a test added for inspecting a code list end-to-end too.
Approved.

CharlesRendle (dismissed):
Requesting changes as ive just seen the build fail

Issue #440, opened 2022-04-21
Find a replacement for shpynx which forces adequate docstring coverage


canwaf:

robons:
TODO: @CharlesRendle is going to write a few short paragraphs describing the key features and differences between the docstring coverage tools - including their key features + differences, how actively maintained they are, how easy they are to use, and whether there are any big outstanding bugs that never get fixed.

PR #785, opened 2023-04-19
#440 build and publis api docs

nimshi89:
#440


source/conf.py
```diff
@@ -0,0 +1,34 @@
...
+# For the full list of built-in configuration values, see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
+
+# -- Project information -----------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+
+project = "csvcubed"
+copyright = "2023, ONS"
+author = "ONS"
+release = "0.3.5"
```
robons, line 12: This needs to be automatically updated when we do a release (via one of the GitHub workflows)

nimshi89: yeah okay will sort it out

Makefile
```diff
@@ -0,0 +1,21 @@
...
+# Put it first so that "make" without argument is like "make help".
+help:
+	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+.PHONY: help Makefile
+
+# Catch-all target: route all unknown targets to Sphinx using the new
+# "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
+%: Makefile
+	echo "Found $@"
```
robons, line 20: We should get rid of this unnecessary line now.

SarahJohnsonONS:
Nick and I had a look, but we're struggling to understand the step-by-step process for generating the docs (e.g. where to set specific modules to be included in the documentation). May be useful to have a chat with Charles/Jozsef so we can understand the process.

canwaf (approved):
Good comments and a lot of fixing Sphinx.

Issue #781, opened 2023-04-06
Move the `csvw pull` command from csvcubed-pmd to csvcubed 


robons:
the `csvcubed-pmd` package as a command accessible at `pmdutils csvw pull <path-or-uri-to-csv-w>` which allows us to pull/download a CSV-W built by csvcubed (and all of its dependencies). I frequently find myself wanting to use this, but we hid it away in the csvcubed-PMD library because that's where it's of primary use.

This task is to: Move this pull command into the csvcubed library. The command should be removed from the `csvcubed-pmd` package. This may mean that we're able to remove the dependency that it has on csvcubed. 

- [x] Copy the csvw pull code from csvcubed-pmd package to the main csvcubed package.
- [x] Release csvcubed
- [x] Alter the pmdutils docker image build script so that it automatically installs the latest version of csvcubed inside the docker container.
- [x] Create an updated release for csvcubed-pmd.
- [x] Alter the [csv2pmd pipeline in pmd-jenkins-library](https://github.com/GSS-Cogs/pmd-jenkins-library/blob/csvwlib-build/vars/csvw2pmd.groovy) so that it uses csvcubed to pull CSV-W instead of pmdutils.
- [x] ~Add a tests to [the relevant file](https://github.com/GSS-Cogs/pmd-jenkins-library/blob/csvwlib-build/test/integration/groovy/uk/org/floop/jenkins_pmd/Csvw2PmdPipelineTests.groovy) to check the csvw2pmd pipeline still works now when pulling a CSV-W.~ Turns out there is already a suitable test in there.
- [x] Remove the `pmdutils csvw pull` command (but leave the `pmdutils csvw find-where` command in place).
- [x] Remove the dependency from csvcubed-pmd to csvcubed.
- [x] Release updated csvcubed-pmd without the `pmdutils csvw pull` command in it.
- [x] Update the [dependency diagram miro board](https://miro.com/app/board/uXjVPeHESx8=/) so it's clear that csvcubed-pmd no longer depends on the csvcubed package directly.

PR #786, opened 2023-04-20
Issue #781 - Bringing CSV-W pull command over from pmdutils to csvcubed.

robons:
I've brought the `pull` command over from pmdutils into csvcubed in this ticket. 

I've done some little bits of tidying up, but I haven't dramatically changed the structure at all. The one thing that I have done is gone through and added debug/info logging at appropriate locations, so we can help diagnose faults a bit better.

Partially satisfies issue #781.


.github/workflows/reusable-test.yaml
```diff
@@ -58,7 +58,7 @@ jobs:
           key: ${{ inputs.os }}-${{ inputs.python-version }}-poetry-${{ hashFiles('poetry.lock') }}
 
       - name: Install python packages
-        run: poetry install
+        run: poetry install --sync
```
robons, line 61: This is to make sure we remove any packages that poetry doesn't think should be there anymore.

src/csvcubed/cli/csvw/pull.py
```diff
@@ -0,0 +1,213 @@
+"""
```
robons, line 1: For reference, [this](https://github.com/GSS-Cogs/csvcubed-pmd/blob/main/src/csvcubedpmd/csvwcli/pull.py) is where I took this code from.

tests/behaviour/csvw-pull.feature
```diff
@@ -0,0 +1,139 @@
+Feature: Testing the CSV-W pull command in the CLI
```
robons, line 1: For reference, [this](https://github.com/GSS-Cogs/csvcubed-pmd/blob/main/tests/features/csvwcommandline.feature) is the file these behaviour scenarios came from.

Since the `find-where` command isn't being moved over, those scenarios haven't transfered.

src/csvcubed/cli/pullcsvw/pull.py
```diff
@@ -0,0 +1,218 @@
...
+
+    base_url = _get_context_base_url(table_group.get("@context"))
+    if base_url is None:
+        base_url = metadata_file_url
+    elif not looks_like_uri(base_url):
+        base_url = urljoin(metadata_file_url, base_url)
+    _logger.debug("Absolute base URL for document: '%s'", base_url)
+
+    dependencies = set(_get_csv_w_spec_dependencies(table_group))
+    dependencies |= _get_rdf_file_dependencies(metadata_file_url)
```
robons, line 112: This operation here is to union the values of the two sets together.

NickPapONS (approved):
Had a look at what was moved over and although I am far from knowledgeable on this command, it all looks good to me. I really like the logging, it's very thorough.

Has approval from me, but I do suggest at least one other person has a look.

Issue #782, opened 2023-04-11
Integrate prototype CsvWBrowser into csvcubed


robons:
Based on a quick play around with the existing Inspect API (i.e. the Inspector classes), it has become apparent that it's really difficult to use the inspector classes to get the information you need out of the CSV-W. This is because the inspector classes are designed to extract data in a performant way and so is aligned with the way data is stored under the hood. This ticket is to adapt this API into an interface which is more user-friendly making it possible to 'browse' the CSV-W  abit more naturally.

- [x] Take the [prototype](https://gist.github.com/robons/86c2c0b3767f2849849049ee4c9400a5) code developed by @robons and bring it into the csvcubed code base.
   - This should be in a new `csvcubed.inspect` package.
   - [x] Make sure to split the prototype code up into appropriate sub-packages and sub-modules so that it isn't one gigantic mess. Take care to ensure that classes and functions are grouped appropriately, so we don't have unrelated classes/functions/concepts in the same fle.
- [ ] For each of the branching logic parts of code, we need to create unit tests to ensure that we're correctly representing the underlying test-case CSV-W's structure.
  - We can probably create three unifying test cases which cover the majority (if not all) of the different branches of code we have. For example:
      - one test-case might be a pivoted shape cube with a LocalAttriibute, ExternalAttribute, Local Dimension, External Dimension, Pivoted Shaped Obs val column and Suppressed column
      - the second test case could be a Standard shape input with locally defined units, locally defined measures
      - the third test case could be a Standard shape input with externally defined units and externally defined measures.

N.B. We only have to test that the browser/interfaces code extracts the correct value from the underlying Inspector functions, we don't need to test that the underlying inspectors functions cope with all input types. This should limit the number of tests we'll actually need here - but it will make it a bit more confusing as to where you should write you test in future. 

 - [ ] Write some kind of blurb explaining when you should write a test against the CsvWBrowser inteface and what you should be testing, versus when you should write a test against the inspector classes and what you should be testing there.

This ticket is a spike. We expect the code to have been split up and be in a (roughly) working order upon completion of this ticket - i.e. basic tests should be in place for the top-level objects. We don't expect all unit tests to have been written. Upon completion of this spike, any remaining tests should be listed in a follow-up ticket which can be added to the next sprint.

- [ ] See if you can come up with a more specific and helpful name than `CsvWBrowser` since this will only successfully browse the CSV-Ws generated by csvcubed.

PR #792, opened 2023-05-03
#782 integrate prototype csv w browser into csvcubed

SarahJohnsonONS:
The [prototype](https://gist.github.com/robons/86c2c0b3767f2849849049ee4c9400a5) code developed by @robons has been added to a new `csvcubed.inspect` package. It has been split up into appropriate sub-packages and sub-modules as follows:
- `browsers.py`: contains the `TableBrowser` and `MetadataBrowser` classes
- `browsertable.py`: contains the `DataCubeTable`, `CodeListTable` and `CsvWBrowser` classes, and the `_map_to_data_cube_column` function which generates the correct column classes from the input CSV-W (see `browsercolumns.py` for more details)
- `browsercomponents.py`: contains all classes defining components as they relate to CSV-W columns
- `browsercolumns.py`: contains all classes defining the CSV-W columns
- `lazyfuncdescriptor.py`: contains the `LazyFuncFieldDescriptor` class and `lazy_func_field` function associated with it.

Unit tests have been created for two of the suggested test cases: 
- A pivoted shape cube with a Local Attribute, External Attribute, Local Dimension, External Dimension, Pivoted Shape Observations column and Suppressed column
- A standard shape input with locally defined units and locally defined measures

These two unit tests demonstrate how a typical end user could access the various tables and columns and their constituent properties and metadata. They also test that the correct type of column and associated component (Dimension/Attribute/Unit/Measure/Observation) are identified when the `CsvWBrowser` object is instantiated, as well as whether the relevant Dimension/Attribute/Unit/Measure is locally or externally defined.

We haven't added any tests that check that specific ValueErrors are raised, but the majority of the branching logic is being tested (see TODOs below).
 
Main TODO:
The third test case (a standard shape input with externally defined units and externally defined measures) has been created, but is waiting for full `UnitColumn` and `MeasureColumn` implementation before unit tests can be developed.

Other TODOs:
Parent/subPropertyOf value to be added to `LocalDimension`/`LocalAttribute`/`LocalUnit`/`LocalMeasure` classes.

Additional unit tests for `CsvWBrowser`:
This depends how deep we expect users to chain properties together. For example, if you call `_get_columns` on a `DataCubeTable` (via the `.columns` property), you can access columns individually using the OrderedDict keys, and then access the relevant component details, e.g `dimension_col.dimension.dimension_uri` or `attribute_col.attribute.label`. But these could also be accessed via a different property chain e.g. `dimension_col.info.component.property` == `dimension_col.dimension.dimension_uri` (or at least, the strings representing these properties are the same…)

Regarding a better name - CsvCubedBrowser? Or is that a bit unwieldy? CubeBrowser? Although that doesn’t make it clear you can also browse code lists…


src/csvcubed/inspect/browsertable.py
```diff
@@ -0,0 +1,175 @@
...
+
+        return columns
+
+    shape: CubeShape = lazy_func_field(_get_shape)
+    data_set_uri: str = lazy_func_field(_get_dataset_uri)
+    columns: OrderedDict[str, DataCubeColumn] = lazy_func_field(
+        _get_columns, repr=False
+    )
+
+    def __hash__(self):
```
robons, line 54: You two comfortable with what the hash function is for?

src/csvcubed/inspect/lazyfuncdescriptor.py
```diff
@@ -0,0 +1,52 @@
+from dataclasses import field
+from typing import Any, Callable, Generic, Mapping, Optional, Type, TypeVar
+
+TClass = TypeVar("TClass")
+TRet = TypeVar("TRet")
+
+
+class LazyFuncFieldDescriptor(Generic[TClass, TRet]):
+    """
+    https://docs.python.org/3/library/dataclasses.html#descriptor-typed-fields
```
robons, line 10: This could use a better explanation of why it's here.

Also, could you check each of the modules you've created and ensure they all have a docstring at the top of the module according to our [dev standards](https://github.com/GSS-Cogs/csvcubed/blob/main/docs/developer.md#api-documentation)?

tests/unit/inspect/test_browsers.py
```diff
@@ -0,0 +1,266 @@
...
+    assert isinstance(pivoted_obs_column.unit, LocalUnit)
+    assert isinstance(pivoted_obs_column.measure, LocalMeasure)
+
+
+def test_standard_shape_browser_locally_defined_measures_units():
+    """
+    Tests whether a CsvWBrowser correctly represents the expected CSVW structure and
+    correctly extracts underlying values from the inspector functions in a standard shape
+    CSVW input with externally defined measures and units.
+    NOTE this test currently does not test anything that isn't covered by the previous test.
```
robons, line 151: Is this true? It tests the Standard shape observations column, measures and units columns.

src/csvcubed/inspect/browsercomponents.py
```diff
@@ -0,0 +1,112 @@
...
+    attribute_component: QubeComponentResult = field(repr=False)
+
+    def _get_attribute_uri(self) -> str:
+        return self.attribute_component.property
+
+    attribute_uri: str = lazy_func_field(_get_attribute_uri)
+
+
+@dataclass(frozen=True)
+class ExternalAttribute(Attribute):
```
robons, line 51: Should we have separate components for External/Local Attribute Resource/Literals?

We can determine the data type of the data in the column for both Internal/ExternalAttributeLiterals which may be helpful in understanding the data.

robons (dismissed):
Very close to being ready for merging! 

Thanks for your efforts on this one, which I know has been a bit confusing at times (due to the lazy func field stuff).

Could you make sure we have the appropriate docstring comments for every module and class we're adding?

And could you also create some follow up tasks for implementation of the todos?

src/csvcubed/inspect/browsertable.py
SarahJohnsonONS: As in, generally, or in this specific case?

robons: Ideally both.

SarahJohnsonONS: Something to do with checking for equality?

tests/unit/inspect/test_browsers.py
NickPapONS: My mistake, this comment should be in the following test

SarahJohnsonONS:
Follow up ticket created (#796)

src/csvcubed/inspect/browsercolumns.py
```diff
@@ -0,0 +1,197 @@
...
+            return ExternalDimension(dimension_component)
+
+        return LocalDimension(dimension_component)
+
+    dimension: Dimension = lazy_func_field(_get_dimension)
+
+
+@dataclass(frozen=True)
+class AttributeColumn(DataCubeColumn):
+    def _get_component(self) -> QubeComponentResult:
```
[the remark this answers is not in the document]
NickPapONS: Done, the dimension class now uses this.

src/csvcubed/inspect/lazyfuncdescriptor.py
NickPapONS: The docstrings have been added to the top of all the new files we made.

CharlesRendle (approved):
Happy that all of the review comments have been addressed and TODO comments in the code are all related to the subsequent ticket.
Approved!

PR #801, opened 2023-05-12
Amended `owl:sameAs` to `skos:exactMatch` in skoscodelistwriter

SarahJohnsonONS:
Made the required update in `writers/skoscodelistwriter.py` and created codelist schema_v1_2 to use `skos:exactMatch` instead of `owl:sameAs`. Followed the instructions in `readers/codelistconfig/codelist_schema_versions.py` to add a new schema. Some unit and behaviour tests were failing, so added an elif statement in `models/codelistconfig/code_list_config.py` to allow for the new `exact_match` property. Feel like there needs to be some error handling/validation to prevent users from using both properties?


src/csvcubed/schema/codelist-config/v1_2/schema.json
```diff
@@ -0,0 +1,182 @@
+{
+    "$schema": "https://json-schema.org/draft-07/schema",
+    "id": "https://purl.org/csv-cubed/code-list-config/v1.1",
```
nimshi89, line 3: I might be wrong but shouldn't this be v1.1?

nimshi89 (changes requested):
Seems really good just have that one question.

src/csvcubed/schema/codelist-config/v1_2/schema.json
SarahJohnsonONS: Good spot, updated

nimshi89 (approved):

robons (changes requested):
I think we should move to a major (breaking) version change on this one. It'll duplicate a bunch of code, but we'll end up with a clearer conceptual break.

PR #816, opened 2023-05-26
#793 Use `skos:exactMatch` instead of `owl:sameAs` in code lists

NickPapONS:
TODO: Write description


nimshi89:
It looks great. I love the comment that explain each step where it's needed.

CharlesRendle (approved):
Sorry this has taken so long. Approving - all looks good to me!

Issue #811, opened 2023-05-22
BUG - Inspect API cannot load dataframe when CSV-W has suppressed columns 


robons:
When attempting to view the data contained within [anxiety.csv-metadata.json](https://raw.githubusercontent.com/GSS-Cogs/family-ess/5a4bd6d3dc522e41e0a504c1cbc51341f951b6e5/out/anxiety/anxiety.csv-metadata.json). I experience the following buggy behaviour:

```python
from csvcubed.inspect.inspectortable import Inspector

anxiety_inspector = Inspector("~/stuff/anxiety.csv-metadata.json")
main_table = anxiety_inspector.tables[0]

main_table.data_cube_repository.get_dataframe(main_table.csv_url)
Traceback (most recent call last):
  File "/Applications/PyCharm.app/Contents/plugins/python/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
           ^^^^^^
  File "<input>", line 1, in <module>
  File "/Users/HA9WF3CZ/Code/csvcubed/src/csvcubed/utils/sparql_handler/data_cube_repository.py", line 350, in get_dataframe
    self._get_new_category_labels_for_col(
  File "/Users/HA9WF3CZ/Code/csvcubed/src/csvcubed/utils/sparql_handler/data_cube_repository.py", line 381, in _get_new_category_labels_for_col
    raise ValueError(
ValueError: Unhandled column type/configuration - Suppressed, ColumnDefinition(csv_url='anxiety.csv', about_url=None, data_type=None, name='areanm', property_url=None, required=False, suppress_output=True, title='AREANM', value_url=None, virtual=False)
```

- [x] Alter the dataframe loading code so that it is possible to generate a DataFrame from a CSV-W data cube which contains suppressed columns.

Triage: 
* Impact = 4/High - no workarounds exist.
* Probability = 2/Low - no external users are actively using this just yet - but this needs to work come September when @canwaf demos the Inspect functionality at the RSS conference.

Overall score: 8 - Important Bug

## Suggestions

Possibly add an extra if statement block that would handle the suppressed columns case (get_new_category_labels_for_column()). 

## Test

Add a test that test this specific case where there are suppressed columns and successfully passes.

PR #818, opened 2023-06-02
#811 inspect api cannot load dataframe when csv w has suppressed columns

SarahJohnsonONS:
New parameter `include_suppressed_cols` added to the `get_dataframe` function (default True) which allows users to specify whether or not to include Suppressed columns in the returned dataframe. Note that the values in Suppressed columns can't be dereferenced to labels as there's no information about the component type so we can't call the correct dereferencing function.


CharlesRendle (changes requested):
Will be approving but requesting to resolve branch conflicts first.
Liking the test coverage added here!

NickPapONS:
> Will be approving but requesting to resolve branch conflicts first. Liking the test coverage added here!

I believe the conflicts are resolved now!

CharlesRendle (approved):
Approving

Issue #783, opened 2023-04-11
Move remaining pieces of inspect functionality into `csvcubed.inspect` package


robons:
Follows on from the work in #782.

There's a bunch of inspect functionality in the `csvcubed.utils` package which probably makes more sense to live in the new `csvcubed.inspect` package. There's probably some stuff in `csvcubed.cli` that could do with being moved as well.

Before beginning the process of moving things over, please examine the contents of utils and the csvcubed.cli.inspect package and think about if there is anything else that was not identified during refinement that should also be moved over.

We want the cli folder to mostly contain code that calls inspect functionality.

For utils, the code here should already be used throughout different areas of the codebase.

The csvcubed.cli.inspectcsvw folder could potentially only contain inspect.py. Other files in inspectcsvw can likely be moved over to new directory.

The three inspector files in csvcubed.utils.sparql_handler should be moved. sparql.py and sparqlquerymanager would be used where all the other files are being moved, so it should be moved as well.

As for column_component_info.py, maybe there should be a new csvcubed.models.inspect package where this (and others like it) could be moved. The existing sparql folder inside models could also go in the new inspect folder. Also from models: sparqlresults.py and inspectdataframeresults.py could go into the new inspect package.

In tableschema.py, CsvWRdfManager is important for the files that are being moved to a new location. This file should be moved over as well.

csvcubed.utils.csvw is relevant to inspect, but we are not sure if it is being used anywhere. If it is not, see if it can be deleted safely.

- [ ] Move all of the relevant files into the csvcubed.inspect package.
  - [ ] Consider the above, as well as any other files that you feel are relevant and could also be moved over.

- [ ] Remember to resolve any import errors that might be caused by moving everything around.

- [ ] The unit tests folder structure should mirror the new structure in the source code. This will mean some unit test files will likely need to be moved around.

Keep an eye out for how much work actually ends up being done in this ticket, if you find that there are more things to move than expected or if the moving process becomes more complicated than expected, then look at creating a follow-up ticket.

PR #821, opened 2023-06-07
Moving remaining pieces of inspect functionality into `csvcubed.inspect` package

nimshi89:
#783


NickPapONS (commented):
Happy with the changes from what I know about the requirements of the ticket. Looks like you covered and cleaned up all the imports as well. Make sure you check the code smells. The vast majority of them are simply "Define a constant instead of using the literal" for test files which is fine, but there are a couple that may have to be addressed. However, I suspect they are actually just caused by things being moved around rather than any code being changed (sonarcloud will pick previously resolved code smells up as new code smells when they are moved to new files or if the file is renamed).

SarahJohnsonONS:
Looks good to me too - unit and behaviour tests all passing. I think the csvcubed/utils/sparql_handler folder can be deleted.

Issue #824, opened 2023-06-14
Use csvw-check inside csvcubed behaviour tests


robons:
We run csvlint as part of a number of cucumber/behave tests when we check that csvcubed is generating valid outputs. 

csvw-check has been written as a more robust and performant alternative to csvlint and (I believe) is ready for internal use now. It recently gave us the heads-up that we were generating technically invalid CSV-Ws: #822.

In this ticket, we should replace the calls to csvlint with a calll to csvw-check. 

**N.B. as part of this we'll need to provide a no-docker way to call csvw-check so that we can run the same behaviour tests in the windows environment.** So make sure to update the reusable-test.yaml github action file once you've made the relevant code changes so that csvlint is swapped out for csvw-check in the Windows runner.

It's possible that in addition to #822 there may be other invalid parts of our CSV-W outputs that this ticket brings up. If you end up with a hornet's nest of problems, stop this ticket and create new ones to address the issues which have surfaced.

CharlesRendle:
Closing as this has been further refined into two tickets (#850 and #851). Hoping that by removing the estimate we gave we will not be collecting points (yet).

Issue #822, opened 2023-06-13
BUG - csvcubed creates invalid CSV-Ws


robons:
When I run csvw-check against one of our cube outputs which has a locally defined code-list, I find that our CSV-Ws are invalid. We get the following error:

```
$ ./csvwcheck-0.1/bin/csvwcheck -s https://raw.githubusercontent.com/GSS-Cogs/family-ess/main/out/anxiety/anxiety.csv-metadata.json
16:32:16.007 [main] DEBUG csvwcheck.Validator -- A @context node must be defined.
16:32:16.010 [main] DEBUG csvwcheck.Validator -- csvwcheck.normalisation.Utils$NormalisationContext.makeError(Utils.scala:663)
csvwcheck.normalisation.Context$.$anonfun$getBaseUrlAndLanguageFromContext$3(Context.scala:63)
scala.Option.getOrElse(Option.scala:201)
csvwcheck.normalisation.Context$.getBaseUrlAndLanguageFromContext(Context.scala:63)
csvwcheck.normalisation.Utils$.$anonfun$fetchRemoteObjectPropertyNode$2(Utils.scala:681)
scala.util.Either.flatMap(Either.scala:352)
csvwcheck.normalisation.Utils$.fetchRemoteObjectPropertyNode(Utils.scala:680)
csvwcheck.normalisation.TableSchema$.tableSchemaPropertyInternal$1(TableSchema.scala:31)
csvwcheck.normalisation.TableSchema$.$anonfun$normaliseTableSchema$12(TableSchema.scala:51)
csvwcheck.normalisation.Utils$.normaliseJsonProperty(Utils.scala:176)
csvwcheck.normalisation.Utils$.$anonfun$normaliseObjectNode$1(Utils.scala:545)
scala.collection.ArrayOps$.map$extension(ArrayOps.scala:924)
csvwcheck.normalisation.Utils$.normaliseObjectNode(Utils.scala:542)
csvwcheck.normalisation.Table$.$anonfun$normaliseTable$1(Table.scala:32)
csvwcheck.normalisation.TableGroup$.$anonfun$normaliseTables$2(TableGroup.scala:116)
scala.collection.Iterator$$anon$9.next(Iterator.scala:575)
scala.collection.IterableOnceOps.foldLeft(IterableOnce.scala:646)
scala.collection.IterableOnceOps.foldLeft$(IterableOnce.scala:642)
scala.collection.AbstractIterator.foldLeft(Iterator.scala:1279)
[...]

PR #828, opened 2023-06-19
822 context bug

CharlesRendle:
Addresses bug ticket #822

csvw-check was finding an error in our cube outputs when they have a locally defined code list.
`tables.1.tableSchema: A @context node must be defined.`

To solve this issue, we have added an extra line when writing in `skoscodelistwriter.py` and specifically to the `table.json` files so that they begin with a @context node.

This change results in a `Valid CSV-W` message after running csvw-check. However, we wanted to include a test which would demonstrate this change but we are unable to use csvw-check in our behave tests until ticket #824 has been completed. Until then, we have added an extra step to one behave test which verifies that the `skoscodelistwriter` has included the node in the correct place. I have added a comment in the code to explain that this should be deleted when the followup ticket has been completed.

### DIFF
--- a/src/csvcubed/writers/skoscodelistwriter.py
+++ b/src/csvcubed/writers/skoscodelistwriter.py
@@ -175,6 +175,7 @@ class SkosCodeListWriter(WriterBase):
             }
         )
         return {
+            "@context": "http://www.w3.org/ns/csvw",
             "columns": csvw_columns,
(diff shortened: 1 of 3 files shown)


NickPapONS (approved):
Looks good, nicely written comment on the temporary change, it is well documented and makes it clear what should be changed when csvwecheck is implemented in behaviour tests.

PR #839, opened 2023-07-03
adding the pandas call

nimshi89:
There was an error when releasing the new version

https://github.com/GSS-Cogs/csvcubed/actions/runs/5444432557


CharlesRendle (approved):
Happy that this change should fix the error highlighted.

PR #855, opened 2023-07-19
MuazzamChaud/issue_754

MuazzamChaud:


src/csvcubed/models/cube/qb/components/codelist.py
```diff
@@ -113,12 +116,31 @@ def _get_arbitrary_rdf(self) -> List[TripleFragmentBase]:
...
+                            cell_uri_template, {csvw_safe_col_name: c}
+                        ),
+                    )
+                    for c in sorted(set(columnar_data))
+                ],
+                uri_style=uri_style,
+            )
+        else:
+            concepts = [NewQbConcept(c) for c in sorted(set(columnar_data))]
+            # todo: do we want to return a CompositeQbCodeList here?
```
robons, line 142: Yeah, but it's up above.

NickPapONS (commented):
Looks good to me, all I would suggest is take a look at the code smells (mostly about using literals which you can just mark as Won't Fix, there is also one smell about exception types) then it can be approved. And probably rename the pull request when it is merged so it has the ticket title.

src/csvcubed/readers/cubeconfig/v1/columnschema.py
```diff
@@ -90,7 +90,7 @@ def map_to_new_qb_dimension(
             # If so, then the code_list provided is not a boolean value or looks like an uri.
             if not isinstance(self.code_list, bool):
                 if looks_like_uri(str(self.code_list)) == False:
-                    raise BaseException(
+                    raise UserWarning(
```
NickPapONS, line 93: Sorry if I misunderstood this, do we want this condition to terminate the program? Or simply give a warning and continue? If a warning is fine then this looks ready.

NickPapONS (approved):
Looks ready, well done!

Issue #856, opened 2023-07-20
[BUG] Cannot use 'Value' as the column name for observation column


GregoryJPavier:
**Describe the issue:**
If a dataset uses the word 'Value' as a title for the observation column it will build, but when inspecting the output it will result in an error.

If you were to commit and push changes made to a dataset or its json file (where 'Value' is used as above), e.g. total-value-of-uk-exports under the Family-ESS repo, then the github build action will also fail, resulting in the same error in the logs.

**Reproducible steps / Code example:**
Find any dataset with 'Value' as the title for the observation column, build it using its config json file, then inspect the output.

OR

Push changes made to such a dataset or its json file to trigger a build.
The output of a failed build can be seen here: https://github.com/GSS-Cogs/family-ess/actions/runs/5599506508/jobs/10240554713

**Error message obtained / Resulting behaviour:**
`2023-07-20 17:48:14,961 - csvcubed.cli.inspectcsvw.inspect - DEBUG - Metadata json-ld path: /Users/gregorypavier/family-ess/out/total-value-of-uk-exports.csv-metadata.json (inspect.py:33)
2023-07-20 17:48:15,512 - csvcubed.inspect.tableschema - INFO - Successfully parsed csvw json-ld to rdf graph. (tableschema.py:145)
2023-07-20 17:48:15,512 - csvcubed.inspect.sparql_handler.sparqlquerymanager - DEBUG - Root path: /Users/gregorypavier/.pyenv/versions/3.10.10/lib/python3.10/site-packages/csvcubed (sparqlquerymanager.py:101)
2023-07-20 17:48:15,512 - csvcubed.inspect.sparql_handler.sparqlquerymanager - DEBUG - select_csvw_table_schema_file_dependencies query file path: /Users/gregorypavier/.pyenv/versions/3.10.10/lib/python3.10/site-packages/csvcubed/inspect/sparql_handler/sparql_queries/select_csvw_table_schema_file_dependencies.sparql (sparqlquerymanager.py:110)
2023-07-20 17:48:15,656 - csvcubed.inspect.tableschema - DEBUG - Loading dependent file containing table schema year.table.json into RDF graph. (tableschema.py:73)
2023-07-20 17:48:15,657 - csvcubed.utils.json - DEBUG - Loading JSON from file /Users/gregorypavier/family-ess/out/year.table.json (json.py:58)
[...]

PR #857, opened 2023-07-24
[BUG] Cannot use 'Value' as the column name for observation column

CharlesRendle:
Addresses #856 in which when a column is titled `Value`and we need to use the pandas `melt()` function such as when calling `transform_dataset_to_canonical_shape()` which causes an error because pandas does not allow a melted dataframe to have a column named `Value`. 

This ticket adds a rather hacky workaround in which we catch any columns called `Value` and give the melt function a `value_name` of `Not-Value` (which is then used in the melted df). We then rename the `Not-Value` column from the melted df and pass this back.

Added a unit test to verify this functionality. Compares column names from the original dataframe are a subset of the ones in the melted df.

### DIFF
--- a/src/csvcubed/utils/csvdataset.py
+++ b/src/csvcubed/utils/csvdataset.py
@@ -307,12 +307,13 @@ def _melt_data_set(
     id_cols = list(set(data_set.columns) - set(value_cols))
 
     # Checking for any columns with the title "Value" and changing the value_name
-    # parameter passed to the melt function to "Not-Value" so that we don't
+    # parameter passed to the melt function to a random string so that we don't
     # trigger a pandas ValueError.
     value_name = "Value"
     for col_title in value_cols:
         if col_title == "Value":
-            value_name = "Not-Value"
+            rand_value_name = str(uuid1())
+            value_name = rand_value_name
 
     # Melting the data set using pandas melt function.
     melted_df = pd.melt(
@@ -324,8 +325,8 @@ def _melt_data_set(
     )
 
     # Renaming columns in the returned melted df to their original title "Value"
-    if value_name == "Not-Value":
-        melted_df.rename(columns={"Not-Value": "Value"}, inplace=True)
+    if value_name == rand_value_name:
+        melted_df.rename(columns={rand_value_name: "Value"}, inplace=True)
 
     return melted_df
 
--- a/src/csvcubed/utils/csvdataset.py
+++ b/src/csvcubed/utils/csvdataset.py
@@ -310,8 +310,9 @@ def _melt_data_set(
     # parameter passed to the melt function to "Not-Value" so that we don't
     # trigger a pandas ValueError.
     value_name = "Value"
-    if "Value" in value_cols:
-        value_name = "Not-Value"
+    for col_title in value_cols:
+        if col_title == "Value":
+            value_name = "Not-Value"
 
     # Melting the data set using pandas melt function.
     melted_df = pd.melt(
--- a/src/csvcubed/utils/csvdataset.py
+++ b/src/csvcubed/utils/csvdataset.py
@@ -310,9 +310,9 @@ def _melt_data_set(
     # parameter passed to the melt function to a random string so that we don't
     # trigger a pandas ValueError.
     value_name = "Value"
+    rand_value_name = f"Value_{str(uuid1())}"
     for col_title in value_cols:
         if col_title == "Value":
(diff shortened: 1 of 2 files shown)


src/csvcubed/utils/csvdataset.py
```diff
@@ -306,15 +306,28 @@ def _melt_data_set(
     ]
     id_cols = list(set(data_set.columns) - set(value_cols))
 
+    # Checking for any columns with the title "Value" and changing the value_name
+    # parameter passed to the melt function to "Not-Value" so that we don't
+    # trigger a pandas ValueError.
```
NickPapONS, line 311: I like the comment, should avoid confusion in the future if anyone else comes across this.

src/csvcubed/utils/csvdataset.py
```diff
@@ -306,15 +306,28 @@ def _melt_data_set(
...
+    value_name = "Value"
+    if "Value" in value_cols:
```
NickPapONS, line 313: Does the pandas ValueError trigger if the column title contains "Value" (e.g. `if "Value" in`), or only if the title is "Value" exactly (== "Value)? I am assuming it is the latter, since this seems to work and I guess it wouldn't otherwise, but I am just checking because I don't know what the error looks like.

tests/unit/inspect/test_inspectdatasetmanager.py
```diff
@@ -589,6 +599,17 @@ def test_get_val_counts_info_multi_unit_single_measure_dataset():
         / "new"
         / "multi-unit-single-measure-dataset.csv-metadata.json"
     )
+    # csvw_metadata_json_path = (
+    #     _test_case_base_dir
+    #     / "multi-unit_single-measure"
+    #     / "final-uk-greenhouse-gas-emissions-national-statistics-1990-to-2019.csv-metadata.json"
```
NickPapONS, line 605: Some commented out code here.

tests/unit/inspect/test_inspectdatasetmanager.py
```diff
@@ -829,3 +850,45 @@ def test_get_concepts_hierarchy_info_hierarchy_with_depth_more_than_one():
...
     assert result.tree.depth() == 2
     assert len(result.tree.all_nodes_itr()) == 10
+
+
+def test_melt_works_when_column_titled_value_exists():
+    """
+    When a df with a column titled "Value" is passed to the melt function in
+    pandas 2.0 a ValueError is produced because of the column title. This test
+    uses a dataset with a "Value" titled column and should be melted when
+    transform_dataset_to_canonical_shape is called. In this case, the "value_name
```
NickPapONS, line 860: Liking the detail on the docstring!

NickPapONS (changes requested):
We may find it "hacky", but I can't think of a simpler or better way to deal with this very specific bug! Just a couple of questions and something to clean up, then should be good for approval.

src/csvcubed/utils/csvdataset.py
SarahJohnsonONS: I don't think you need to set value_name to "Value", you can just go straight to checking if "Value" is in value_cols

src/csvcubed/utils/csvdataset.py
CharlesRendle: Thanks, this is where code should be removed when addressing the [follow up ticket](https://app.zenhub.com/workspaces/features-squad-61b72275ac896c0010a1b00b/issues/gh/gss-cogs/csvcubed/860) #860 .

src/csvcubed/utils/csvdataset.py
CharlesRendle: @NickPapONS - Changed to loop through column titles in `value_cols` instead of only matching `"Value" in ..`

@SarahJohnsonONS - I think we need to assign a default to `value_name` because its only if `Value` is a title in `value_cols` do we assign a new string to `value_name` but we pass `value_name` regardless to `melt()`

tests/unit/inspect/test_inspectdatasetmanager.py
CharlesRendle: Thanks, good spot

tests/unit/inspect/test_inspectdatasetmanager.py
CharlesRendle: Wondered if it was a bit long winded but I think it will give some context for the follow-up ticket #860

Issue #850, opened 2023-07-13
csvw-check validating linux behaviour tests


CharlesRendle:
Ticket is to replace csvlint checks in linux tests while keeping csvlint in the windows tests (tb-fixed in a follow-up ticket). 

- [x] In `csvlint.py` make a new function that follows the structure of `_run_csvlint()` in that if `SHOULD_USE_DOCKER` is `True` then we pull the csvw-check docker image ([as described in csvw-check readme](https://github.com/GSS-Cogs/csvw-check)) and run the validation from there. *Note* make sure we capture the logs from csvw-check output. If `SHOULD_USE_DOCKER` is false, we are assuming that this indicates we are in the windows environment so for the time being continue to use csvlint (same as `else:` in `_run_csvlint()`. 
<br/>
- [x] Alter (or make new step definitions) in `csvlint.py` so that the new function is being called instead of 'run_csvlint()`.
    - [x] Make sure that step definitions are written for:
    - all CSV-Ws should succeed
    - "{file}" should succeed
    - "{file} should fail with "{expected}" </br>

- [x] Update all feature files which call the original csvlint steps (mostly in `qbwriter.feature` & `skoscodelistwriter.feature` & 1 in `cube-config.feature`) so that they call the new/updated step definitions.
<br/>
- [ ] The file with the existing step definitions is `csvlint.py` this needs to be changed at some point to reflect that we are using csvw-check instead. When changed this will also need to be changed in `devtoolsimports.py`.
<br/>
- [x] Be aware that after updating all the behaviour tests to use the new function (aka csvw-check) this might break many tests as it exposes our existing test cases as not passing the new validations. If this is the case, document all of these breakages and make according tickets. Up to the team to decide to to proceed with order of tickets from here as technically this isn't a breaking change but will mean we are pushing to main something which will expose failed tests.

CharlesRendle:
Democratically giving this a `5` @SarahJohnsonONS @NickPapONS

PR #862, opened 2023-07-27
#850 Csvw-check validating Linux behaviour tests

NickPapONS:
This ticket changes csvcubed's behaviour tests that use csvlint so they now use csvwcheck instead. The new behave steps were added into the file `csvlint.py` in the devtools repository, which is then imported into csvcubed. The steps run the new function `_run_csvwcheck` which uses csvw-check on the given input, captures the logs and any errors. The behaviour tests using csvlint that were altered are in `cube-config.feature`, `qbwriter.feature` and `skoscodelists.feature`.

For now, csvw-check is only being used in behaviour tests in the linux environment, as determined by the environment variable "SHOULD_USE_DOCKER". This is why the functionality to use csvlint still remains in `csvlint.py`. This will be removed and the file will be renamed to represent the new use of csvw-check as part of the next ticket that completes the implementation of csvw-check in the windows environment.

Here is the file with the new behaviour steps and csvw-check function that were added in devtools: https://github.com/GSS-Cogs/csvcubed-devtools/pull/7/files
Thanks to @SarahJohnsonONS for creating the pull request in that repository.

Important Note: One of the behaviour test steps, named `'csvwcheck validation of "{file}" should fail with "{expected}"'`, has been implemented in the devtools, but is not currently being used in the behaviour tests. This is because the csvlint version of this step outputs a different error log message than csvw-check, and the error log is what is being used to assert the result of the test. This means if the csvlint version of the step is replaced with the csvwcheck version at this point, then the Windows tests cannot pass as they are currently still using csvlint and are expecting a different exit error message to assert on. There are 4 behaviour scenarios using this step, so they are the only ones currently still using the old csvlint steps. To avoid breaking tests, I recommend for these 4 scenarios to be replaced with the new csvwcheck step in the follow-up ticket that focuses on making csvw-check work in the Windows tests.


canwaf (approved):
Looks good, only floated devtools version, just in case we want to update it.

PR #878, opened 2023-08-24
831 csvcubed repo distribution implementation

NickPapONS:


dcatDistribution.json
canwaf, line 1: This should be deleted.

dcatDistribution.ttl
canwaf, line 1: This should be deleted too.

canwaf (approved):
If you delete those two files, you may squash & merge. If they remain, and you squash & merge, I will teach you how to revert changes to main so they are written out of existance and you don't want to spend the five hours of me muttering git incantations under my breath cribbed from stack overflow.

Great work!

PR #882, opened 2023-08-30
#860 display warning/error when csv input contains a observation value column titled `value`

SarahJohnsonONS:
New InvalidObsValColTitleException class created and implemented in `csvdataset.py`. External documentation pages updated and purl link added.


canwaf (approved):
Does what it says in the (updated by you) docs. No notes.

PR #887, opened 2023-09-12
pandas~2.0

SarahJohnsonONS:
Version lock pandas~2.0 until #886 complete

### DIFF
--- a/.github/workflows/main-push.yaml
+++ b/.github/workflows/main-push.yaml
@@ -10,6 +10,8 @@ jobs:
       matrix:
         # Test on Windows + Ubuntu on oldest + newest supported versions of python.
         python-version: ['3.9', '3.11']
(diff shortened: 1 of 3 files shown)


NickPapONS (approved):
Approved. As for the errors caused by pandas 2.1.0, I came across many of the same ones locally when working on dcat distribution and had to run poetry update, so I spent some time coming up with fixes. I ended up resolving the errors on my end, but that was only on that branch which already had changes to many other files so I would like to see what your thoughts are on the fixes, see if they make sense? Here is the commit where I was working on the pyright issues which mostly match the ones you listed in #886: https://github.com/GSS-Cogs/csvcubed/pull/878/commits/f0f707bd25f86881d37c6439ce1c2038eef7624f

Let me know what you think, I don't know if my fixes are correct since I was working locally in a branch that already had many other changes in it. But maybe try them on your end and see if it helps with pyright at all.

Issue #888, opened 2023-09-15
Observation column named Value


canwaf:
Right now we prevent the creation of qubes from pivoted csv which has a column called value, and we check columns before we pivot for such names.

But in standard shape we accept the name value, but inspect will fail because we check for the column name before we pivot.

We should just say no. No columns called values from the word go.

(And fix the docs as the link to the docs don't work.)

https://gss-cogs.github.io/csvcubed-docs/external/guides/errors/inspect-command-errors/invalid-obs-val-col-entitled-value/ resolves from http://purl.org/csv-cubed/err/obs-val-col-entitled-value

PR #892, opened 2023-09-26
#888 observation col named value

NickPapONS:
Changes the error condition to be more strict so no observation value columns called "Value" are allowed at all. Any occurrence of "Value" named columns was altered to "Val" or something equivalent to abide by the new error condition.


src/csvcubed/utils/csvdataset.py
canwaf, line 1: Don't comment out deprecated code, just delete it.

canwaf (changes requested):
You also need to clear out the option of "Value" from the [Conventional config approach](https://github.com/GSS-Cogs/csvcubed/blob/156830283aa81e663c081c1c9c8c507c9cda932d/src/csvcubed/readers/cubeconfig/v1/configdeserialiser.py#L305), and update the docs https://gss-cogs.github.io/csvcubed-docs/external/guides/configuration/convention/#conventional-column-names

canwaf (approved):
Good work, and of course we broke our tests.

